PR #13, opened 2017-06-07
DnD fixes

mkoty:


src/common/dnd/drag.directive.ts
```diff
@@ -33,13 +33,13 @@ export class DragDirective implements OnInit, OnDestroy {
 
   start(e) {
     const transfer = e.dataTransfer;
-    if (this.canDrag(this.event()) === false) {
+    if (this.canDrag === false) {
```
klumba12: Shouldn't be it a function?

klumba12: is'it bindined automatically?

src/main/core/head/head-core.component.html
```diff
@@ -6,9 +6,9 @@
       [attr.colspan]="$column.colspan"
       [q-grid-drag]="$view.head.transfer($column.model)"
       [q-grid-drop]="$view.head.transfer($column.model)"
-      q-grid-on-drop="$view.head.drop.execute($event)"
-      q-grid-can-drop="$view.head.drop.canExecute($event)"
-      q-grid-can-drag="$view.head.drag.canExecute($event)"
+      (q-grid-on-drop)="$view.head.drop.execute($event)"
+      [q-grid-can-drop]="$view.head.drop.canExecute($column.model)"
```
klumba12: I thinks this not very good, cause it'll trigger every time on change tracker we need to pass it as a function

core/head/head.view.js
```diff
@@ -12,13 +12,9 @@ export class HeadView extends View {
 
 		this.drop = new Command({
 			canExecute: e => {
-				if (e.source && e.source.key === tagName) {
```
klumba12: Don't we need to check the tag name? It's kinda of convension to check if source of drag/drop is valid

src/common/dnd/drag.directive.ts
mkoty: from head.view.js:
canDrag = `map.hasOwnProperty(e.key)` && map[e.key].canMove !== false`
It's just boolean value

core/head/head.view.js
mkoty: In that moment: `e.source:"user"`

src/common/dnd/drop.directive.ts
```diff
@@ -27,6 +29,7 @@ export class DropDirective implements OnInit, OnDestroy {
     this.listener.on('dragover', this.over);
     this.listener.on('dragleave', this.leave);
     this.listener.on('drop', this.drop);
+    this.canDrop = this.canDropFunction(this.model);
```
klumba12: i think this should be dynamic, depending on dragging stuff so it should be a function

src/common/dnd/drag.directive.ts
klumba12: It can be dynamic depending on grid state so it should be a function

PR #38, opened 2017-08-18
Cell editor for Date column type fixed

azkurban:


azkurban:
previous commits issues have been fixed. 
Added new commits: fix for edit-cell-bool template

PR #39, opened 2017-08-25
Cell editor for Array type implemented

azkurban:


.gitignore
```diff
@@ -73,3 +73,5 @@ dll/
 *.css.shim.ts
 *.ngsummary.json
 *.shim.ngstyle.ts
+src/themes/material/index.css.map
```
klumba12: exclude all files * .css.map

package.json
```diff
@@ -85,6 +85,9 @@
     "@angular/platform-browser": "4.3.4",
```
klumba12: 1. remove package-lock file
2. put all font deps under the devDependencies

src/common/focus/blur.directive.ts
```diff
@@ -17,7 +18,7 @@ export class BlurDirective implements OnInit {
 	ngOnInit() {
 		this.inputElement = this.renderer.selectRootElement('input');
```
klumba12: u dont need to store inputElement in "this" context

src/common/focus/blur.directive.ts
```diff
@@ -17,7 +18,7 @@ export class BlurDirective implements OnInit {
...
 		if (!this.inputElement) {
-			throw Error('Required Input elemement is not found');
+			throw new AppError('blur.directive', 'Required Input elemement is not found');
```
klumba12: Ex msg: input if not found for 'element.name' or something like that

src/themes/material/assets/cell.edit.scss
```diff
@@ -123,9 +123,30 @@ td.q-grid-edit {
 		width: 100%;
```
klumba12: please if you need to change something for mat, add new file with "mat" prefix(mat.cell.edit in this case)

src/themes/material/components/chip-list.component.ts
```diff
@@ -0,0 +1,122 @@
+import {Component, Input, OnInit, ChangeDetectorRef} from '@angular/core';
+import {AppError} from '@grid/core/infrastructure';
+
+interface IChipValue {
+	value: any;
+	index: number;
+}
+const enterKey = 13;
```
klumba12: u can try to use Shortcut service for that

src/themes/material/components/chip-list.component.ts
```diff
@@ -0,0 +1,122 @@
...
+	constructor(private cdRef: ChangeDetectorRef) {
+		this._editMode = false;
+	}
+
+	@Input('cell-value') chipValues: any[];
+
+	private _values: IChipValue[];
+
+	private newValue: any;
+	private _editMode: boolean;
```
klumba12: by our code convension we don't use "_" prefix

src/themes/material/templates/edit-cell-array.tpl.html
```diff
@@ -2,9 +2,7 @@
 	<div class="q-grid-editor q-grid-array">
 		<h2 class="md-title">Edit {{$cell.column.title}}</h2>
 		<div class="q-grid-array-view">
-			<md-chip-list>
-				<md-chip *ngFor="let item of $cell.value" selected="true">{{item}}</md-chip>
-			</md-chip-list>
+			<q-grid-chips [cell-value]="$view.edit.cell.value"></q-grid-chips>
```
klumba12: cell-value to ng-model or items? chips should be independed from cell, even for names

src/themes/material/components/chip-list.component.ts
```diff
@@ -0,0 +1,122 @@
...
+			};
+		});
+	};
+
+	private editMode(index: number) {
+		return this._editMode && this.editIndex === index;
+	}
+
+	private removeItem(index: number) {
+		this.chipValues.splice(index, 1);
```
klumba12: please for remove, add etc. use Command pattern + shortcuts (u can find examples in grid)

src/themes/material/components/chip-list.component.ts
```diff
@@ -0,0 +1,203 @@
...
+		})
+	};
+
+	constructor(private cdRef: ChangeDetectorRef) {
+		const manager = new CommandManager();
+
+		const shortcutCommands = [this.commandSet.endEdit, this.commandSet.addItem];
+		const otherCommands = [this.commandSet.endEdit, this.commandSet.addItem, this.commandSet.deleteItem];
+
+		this.shortcut.register(manager, shortcutCommands);
```
klumba12: does it matter if command has shortcut or not? can we not split commands to this groups?

src/themes/material/components/chip-list.component.ts
```diff
@@ -0,0 +1,203 @@
...
+	set chipList(values: any[]) {
+		this.chipValueMap = values.map((itemValue, itemIndex) => {
+			return <IChipValue> {
+				value: itemValue,
+				index: itemIndex
+			};
+		});
+	};
+
+	private commandSet: ICommandSet = {
```
klumba12: why we need this commandSet abstraction?

src/themes/material/components/chip-list.component.ts
```diff
@@ -0,0 +1,203 @@
...
+
+@Component({
+	selector: 'q-grid-chips',
+	templateUrl: './chip-list.tpl.html'
+})
+export class ChipListComponent implements OnInit {
+	@Input('values') chipValues: any[];
+	private shortcut = new Shortcut(new ShortcutManager());
+
+	private chipValueMap: IChipValue[];
```
klumba12: why we need this kind of abstraction, and can't wokr directly with chipValues?

src/themes/material/components/chip-list.component.ts
```diff
@@ -0,0 +1,203 @@
...
+			if (value === item) {
+				found = true;
+			}
+		});
+		return found;
+	}
+
+	private beginEdit(itemIndex: number) {
+		this.setSelectedAction(Actions.beginEdit, itemIndex);
+		if (this.commandSet.beginEdit.canExecute()) {
```
klumba12: All this method with setAction are looks to heavy, maybe u have some other ideas?

azkurban:
- refactoring performed;
- edit cell array feature: 
  - has been redesigned.
  - added shortcuts support for chip manipulation commands:
Navigation:
    - focus left, focus right for all items (keyboard: 'left', 'right');
    - focus left for empty add new item input (keyboard: 'backspace');
Modification:
- delete (keyboard: 'backspace');
- add new/save edit value (keyboard: 'enter');

src/common/dom-helpers/dom-base.ts
```diff
@@ -0,0 +1,27 @@
+import {Renderer2, ElementRef} from '@angular/core';
```
klumba12: try to avoid `-helper`, `-base` and other meaningless prefixes

src/common/dom-helpers/dom-base.ts
```diff
@@ -0,0 +1,27 @@
...
+
+	protected listen(selector: string | any, event: string, onEvent: Function) {
+		let element = this.selectRootElement(selector);
+
+		this.renderer.listen(element, event, () => {
+			setTimeout(onEvent, 200);
+		});
+	}
+
+	protected selectRootElement(selector: string | any) {
```
klumba12: why `Root`? also when see method like select or find u don't expect exception,
consider to rename it to `getElement(selector)`

src/common/dom-helpers/dom-base.ts
```diff
@@ -0,0 +1,27 @@
...
+		let nativeElement = this.elementRef.nativeElement;
+		let element = this.renderer.selectRootElement(selector);
+		if (!element) {
+			throw new AppError('dom-base', `Required elemement is not found ` +
```
klumba12: format: make it one line or 2lines 
'dom-
`Required-

src/common/dom-helpers/dom-helpers.module.ts
```diff
@@ -0,0 +1,15 @@
+import {NgModule} from '@angular/core';
```
klumba12: blur directive should be in focus folder, we try to have feature-based folder structure

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
```diff
@@ -0,0 +1,198 @@
...
+interface ISelected {
+	index?: number;
+	value?: string;
+	action: Actions;
+}
+
+interface IChipItem {
+	value?: string | any;
+	state?: string;
+	qGridFocusEnabled?: boolean;
```
klumba12: avoid to use qGrid prefix

src/themes/material/templates/edit-cell-bool.tpl.html
```diff
@@ -3,7 +3,7 @@
 		<md-checkbox
 			q-grid-focus="input"
 			[(ngModel)]="$view.edit.cell.value"
-			[q-grid-blur]="$view.edit.cell.commit.execute">
+			[input-on-blur]="$view.edit.cell.commit.execute">
```
klumba12: all common directives and components should start with `q-grid`

src/themes/material/templates/edit-cell-array.tpl.html
```diff
@@ -2,9 +2,7 @@
 	<div class="q-grid-editor q-grid-array">
 		<h2 class="md-title">Edit {{$cell.column.title}}</h2>
 		<div class="q-grid-array-view">
-			<md-chip-list>
-				<md-chip *ngFor="let item of $cell.value" selected="true">{{item}}</md-chip>
-			</md-chip-list>
+			<q-grid-chip-list [values]="$view.edit.cell.value"></q-grid-chip-list>
```
klumba12: just curious, if ngModel is common directive for such cases or not? (instead values)

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
```diff
@@ -0,0 +1,198 @@
...
+	add,
+	delete,
+	focusNext,
+	focusPrevious,
+	inactive,
+}
+
+@Component({
+	selector: 'q-grid-chip-list',
+	templateUrl: './edit-chip-list.tpl.html',
```
klumba12: why edit-chip-list and not just chip-list?

src/themes/material/components/edit-chip-list/edit-chip/edit-chip.component.ts
```diff
@@ -0,0 +1,215 @@
...
+			this.endEditCommand,
+			this.focusNextCommand,
+			this.focusPreviousCommand,
+			this.focusPreviousFromAddNewInputCommand
+		];
+
+		this.shortcut.register(manager, shortcutCommands);
+	}
+
+	private deleteThis() {
```
klumba12: bad name, try to make delete(chip), This - doesn't make any sence

src/themes/material/components/edit-chip-list/edit-chip/edit-chip.component.ts
```diff
@@ -0,0 +1,215 @@
...
+		canExecute: () => this.nextAction === NextAction.endEdit && this.editValue,
+		execute: () => {
+			this.save.emit(this.editValue);
+			this.endEdit();
+			this.editValue = '';
+		},
+		shortcut: 'enter'
+	});
+
+	private addItemCommand = new Command({
```
klumba12: avoid Command prefix

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
```diff
@@ -0,0 +1,198 @@
...
+} from '@angular/core';
+
+import {AppError} from 'ng2-qgrid/core/infrastructure';
+import {Command, CommandManager} from 'ng2-qgrid/core/command';
+import {Shortcut, ShortcutManager} from 'ng2-qgrid/core/shortcut';
+import {EditChipComponent} from './edit-chip/edit-chip.component';
+import {ChipState} from './edit-chip/edit-chip.component';
+
+interface ISelected {
+	index?: number;
```
klumba12: why u need this abstraction?

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
```diff
@@ -0,0 +1,198 @@
...
+	value?: string;
+	action: Actions;
+}
+
+interface IChipItem {
+	value?: string | any;
```
klumba12: why u need this abstraction?

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
```diff
@@ -0,0 +1,198 @@
...
+	state?: string;
+	qGridFocusEnabled?: boolean;
+}
+
+enum Actions {
+	add,
```
klumba12: explain please why u need this?

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
```diff
@@ -0,0 +1,198 @@
...
+	private items: IChipItem[];
+
+	private shortcut = new Shortcut(new ShortcutManager());
+
+	private selectedItem: ISelected = {
+		action: Actions.inactive
+	};
+
+	private incrementSelectedIndex() {
+		if (this.selectedItem.index === this.chipValues.length) {
```
klumba12: bad names and too complicated, selectedItem.index is a confusion part, too much state in one variable

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	private saveItem(value: string | any, index: number) {
+		if (this.saveItemValueCommand.canExecute(value, index)) {
+			this.saveItemValueCommand.execute(value, index);
+		}
+	}
+
+	private deleteChip(itemIndex: number) {
+		this.selectedItem = {
```
klumba12: try to get rid of all this wrappers for the commands: this.deleteItem.execute(index) looks clear

src/common/dom-helpers/input-blur.directive.ts
```diff
@@ -0,0 +1,20 @@
+import {Directive, OnInit, Input, Renderer2, ElementRef} from '@angular/core';
```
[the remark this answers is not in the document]
azkurban: I've named it so because of the selector: that is the hardcoded 'input' in this case:
`ngOnInit() {
 +		this.listen('input', 'blur', this.onBlur);
 +	}`
And before refactoring the code was like this:
`this.inputElement = this.renderer.selectRootElement('input');
...
this.renderer.listen(this.inputElement, 'blur', () => {
 -			setTimeout(this.onBlur, 200);`

To get rid of the 'input' prefix we need to rewrite the code, so the directive selector receives the value from the argument dynamically instead of being hardcoded input.

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: This abstraction is represented a selected chip object, that needs to store a few property values. I think it is a better option rather than store the properties of a single object (obviously, it is an object by its nature) in a few independent variables and manipulated with them in a designated function in one place like
 ` updateSelectedValues(index, value, action) `
 (although this is an option as well). This would be more like procedural programming and seems to make the code harder to understand. But maybe I am wrong?

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: The answer is similar to my previous comment. Furthermore, it seems to be harder to implement required functionality without this abstraction, considerably increasing the code complexity and reducing its elegance.
Secondly, I believe I will need this abstraction to add the TrackBy support to the component because it will be necessary to track additional property values of the item, not just the value to be rendered on the form.

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: I meant that this list is editable, not just read only view.

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: Yes, it might be this enum is not needed in this component. But it seems that the Command pattern (where I used the enum values to decide what the action expected at the moment) is not needed in this case as well?
So I can refactor the code, to get rid both of this enum and unnecessary Command pattern using. How's that?

src/common/dom-helpers/dom-base.ts
azkurban: according to this remark, files have been renamed in the commit: [blur directive, dom base class: refactoring](https://github.com/qgrid/ng2/pull/39/commits/bb9b43a5374e93ada17ace5c6f028fd5aa7ac651).

src/common/dom-helpers/dom-base.ts
azkurban: according to this remark, the method has been renamed to 'getElement' in the commit: [blur directive, dom base class: refactoring](https://github.com/qgrid/ng2/pull/39/commits/bb9b43a5374e93ada17ace5c6f028fd5aa7ac651).

src/common/dom-helpers/dom-base.ts
azkurban: according to another remark of this review, the code throwing the AppError has been removed in the commit: [blur directive, dom base class: refactoring](https://github.com/qgrid/ng2/pull/39/commits/bb9b43a5374e93ada17ace5c6f028fd5aa7ac651).

src/common/dom-helpers/dom-helpers.module.ts
azkurban: according to this remark, the blur directive has been moved into the 'focus' folder in the commit: [blur directive, dom base class: refactoring](https://github.com/qgrid/ng2/pull/39/commits/bb9b43a5374e93ada17ace5c6f028fd5aa7ac651).

src/common/dom-helpers/input-blur.directive.ts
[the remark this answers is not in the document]
azkurban: according to this remark, the 'input-blur directive' has been renamed to 'blur directive' in the commit: [blur directive, dom base class: refactoring](https://github.com/qgrid/ng2/pull/39/commits/bb9b43a5374e93ada17ace5c6f028fd5aa7ac651).
The directive has been refactored, so it now receives the input property value through the 'q-grid-on-blur-selector' attribute that has the 'input' value by default

src/themes/material/templates/edit-cell-bool.tpl.html
azkurban: according to this remark, the directive has been renamed to 'q-grid-on-blur' in the commit: [blur directive, dom base class: refactoring](https://github.com/qgrid/ng2/pull/39/commits/bb9b43a5374e93ada17ace5c6f028fd5aa7ac651).

src/themes/material/components/edit-chip-list/edit-chip/edit-chip.component.ts
azkurban: Do you mean 'Command' **postfix**?

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: the code has been refactored in commit [refactoring: qGridFocus... properties renamed](https://github.com/qgrid/ng2/pull/39/commits/46b594341d4c95229d6419f373228890b2adefa5).
The properties 'qGridFocus(Enabled/Disabled)' has been renamed to 'focusDirective(Enabled/Disabled)'.

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: Do I understand you correctly: I needn't use here the pattern like:
```
                if (this.addItem.canExecute(value)) {
			this.addItem.execute(value);
		}
```
This is an atomic block of code that means a single action: addItem, so it looks like make sense to wrap it into method: addItem(). Also, most of this wrapper methods are to be embedded into the component template. It would be rather inconvenient to embed into template all this peace of code above

azkurban: So another option would be if I don't use 'canExecute()' method at all and check all conditions needed in the 'execute() method instead. Then I wouldn't need a wrapper and can just call addItem.execute() method and painlessly embed this call into the template. So, what do you think about all this?

src/common/dom/dom.ts
```diff
@@ -0,0 +1,21 @@
+import {Renderer2, ElementRef} from '@angular/core';
+import {AppError} from 'ng2-qgrid/core/infrastructure';
+
+export class Dom {
```
klumba12: Bad name too common

klumba12: In my opinion this is redunded abstraction, if you don't wont to avoid copy paste by this, please make refactor it to composition(now is deriviation). Also be free to inject this code explicitly where it need.

src/common/focus/focus.module.ts
```diff
@@ -1,18 +1,15 @@
 import {NgModule} from '@angular/core';
 import {FocusDirective} from './focus.directive';
 import {AutoFocusDirective} from './autofocus.directive';
-import {BlurDirective} from './blur.directive';
```
klumba12: blurdirective should be in this module

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
klumba12: i'm not sure if u need `updateSelectedValues(index, value, action)` i agree that should be selectionIndex property all other state looks redundant from mine point of view.

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
klumba12: there can be `placeholder` abstraction to create new values but don't understand whe we need to wrap chip values, what state can have particular chip? 
consider for next composition:
1. one state for whole chip-list = 'edi | view'
2. selectionIndex property for whole chip-list (selectionIndex === -1 means placeholder is active)

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
klumba12: i think this is unnessarry, we wont add prefixes for each state chip, common name `chip-list` is totally ok, an

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
klumba12: the action in ui should be disabled if canExecute === false, so please use it in template instead of ts code. Imagine that u are creating API for other people, it will be confusing why we have delete.execute and deleteChip together, wont it?

src/themes/material/components/edit-chip-list/edit-chip/edit-chip.component.ts
klumba12: just think about different names - no suffix

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: Regarding the TrackBy support: if I understand correctly this feature requires identifying the item so it can track its state changes. Just plain value is not enough as we agreed that is isn't necessarily unique. So, using this abstraction I can use the item index as an ID and track other property values, such as value, selected state, mode (view/edit) when they are changed

src/themes/material/templates/edit-cell-array.tpl.html
azkurban: Agree, this is a good idea to use ngModel in this case. Yet, ngModel support requires additional implementation work. I am going to implement this after I've done with other remarks in this review

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: FYI: the following are some of the useful resources I used to familiarize myself with the Track By:
- [How to use trackBy with ngFor](https://stackoverflow.com/questions/42108217/how-to-use-trackby-with-ngfor);
- [how to use track by inside ngFor angular 2](https://stackoverflow.com/questions/36322829/how-to-use-track-by-inside-ngfor-angular-2).

src/common/focus/focus.module.ts
azkurban: Done in the commit: [Merge from Dev: conflicts resolving](https://github.com/qgrid/ng2/pull/39/commits/da05f15a228a611a5059e3049c9f18402446fa8f)

src/themes/material/assets/cell.edit.scss
azkurban: Done

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: I've removed this abstraction after refactoring in the commit: [Merge from Dev: conflicts resolving](https://github.com/qgrid/ng2/pull/39/commits/da05f15a228a611a5059e3049c9f18402446fa8f).
Moreover, since the IChipItem abstraction still remains there was no need to use selectionIndex property. So it wasn't added.

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: Removed this enum, (as well as NextAction enum from chip.component.ts)
(commit: [Merge from Dev: conflicts resolving](https://github.com/qgrid/ng2/pull/39/commits/da05f15a228a611a5059e3049c9f18402446fa8f)).

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: Done, during refactoring process in the commit: [Merge from Dev: conflicts resolving](https://github.com/qgrid/ng2/pull/39/commits/da05f15a228a611a5059e3049c9f18402446fa8f).

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: Have refactored this in the commit: [Merge from Dev: conflicts resolving](https://github.com/qgrid/ng2/pull/39/commits/da05f15a228a611a5059e3049c9f18402446fa8f).

src/themes/material/components/edit-chip-list/edit-chip-list.component.ts
azkurban: Refactored this in the commit: [Merge from Dev: conflicts resolving](https://github.com/qgrid/ng2/pull/39/commits/da05f15a228a611a5059e3049c9f18402446fa8f).

src/themes/material/components/edit-chip-list/edit-chip/edit-chip.component.ts
azkurban: Refactored this in the commit: [Merge from Dev: conflicts resolving](https://github.com/qgrid/ng2/pull/39/commits/da05f15a228a611a5059e3049c9f18402446fa8f).

src/themes/material/components/edit-chip-list/edit-chip/edit-chip.component.ts
azkurban: Refactored this in the commit: [Merge from Dev: conflicts resolving](https://github.com/qgrid/ng2/pull/39/commits/da05f15a228a611a5059e3049c9f18402446fa8f).

PR #53, opened 2017-09-25
Scrollbar

munkk:


package.json
```diff
@@ -90,6 +90,8 @@
     "@angular/platform-browser": "~4.3.4",
     "@angular/platform-browser-dynamic": "~4.3.4",
     "@angular/platform-server": "~4.3.4",
+    "ngx-perfect-scrollbar": "^4.6.5",
+    "perfect-scrollbar": "^0.8.0",
```
titov-max: Why do you need to install this package too?

src/main/core/table/table-core.component.html
```diff
@@ -6,6 +6,7 @@
 			 q-grid-core-head>
 	</thead>
 	<tbody *ngIf="visibility.body"
+			 [perfect-scrollbar]="config"
```
titov-max: I think you can write just ```[perfect-scrollbar]``` while config is empty.

demo/index.html
```diff
@@ -21,6 +21,7 @@
 	<!--<link href="../node_modules/@angular/material/prebuilt-themes/indigo-pink.css" rel="stylesheet">-->
 	<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
 	<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:400,700">
+	<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jquery.perfect-scrollbar/0.7.0/css/perfect-scrollbar.min.css" />
```
titov-max: I guess css should be added somewhere in main app, not demo. User shouldn't think about import of these styles.

src/main/core/table/table-core.component.html
titov-max: And remove ```config``` property from component. Now it looks redundant.

PR #67, opened 2017-10-03
Scrollbar.

munkk:


src/assets/scrollbar.scss
```diff
@@ -0,0 +1,139 @@
+//colors
```
klumba12: please add this under template/material folder

klumba12: is it system? why we can't just include it from node_modules?

src/assets/view.scss
```diff
@@ -13,6 +13,10 @@
 	> .q-grid-table-left {
```
klumba12: put this under material folder

src/main/core/scrollbar/scrollbar.directive.ts
```diff
@@ -0,0 +1,41 @@
+import {AfterViewInit, Directive, ElementRef, Input, Output, OnInit, HostListener} from '@angular/core';
+import {PerfectScrollbarConfigInterface} from 'ngx-perfect-scrollbar';
+import {PerfectScrollbarDirective} from 'ngx-perfect-scrollbar';
+import {EventListener} from 'ng2-qgrid/core/infrastructure/event.listener';
+import {EventManager} from 'ng2-qgrid/core/infrastructure/event.manager';
+
+@Directive({
+	selector: '[q-grid-scrollbar]'
```
klumba12: i can't see directive is uded where is that?

PR #71, opened 2017-10-16
id template

munkk:


src/themes/material/templates/body-cell-row-number.tpl.html
```diff
@@ -0,0 +1,3 @@
+<ng-template key='body-cell-row-number.tpl.html' let-$cell>
```
klumba12: you need to include this file to index.ts

PR #94, opened 2017-10-30
Image type column (view mode)

azkurban:
Only view mode implemented


demo/app/home/home.component.ts
```diff
@@ -35,11 +53,27 @@ export class HomeComponent implements OnInit {
...
 			.map(humans => this.madeIsFeemaleField(humans))
 			.map(humans => this.madeEmailSingleField(humans))
+			.map(humans => this.madeAttachementField(humans))
+			.map(humans => this.madeAvatarField(humans))
 			.subscribe(people => {
 				this.rows = people;
 			});
 	}
 
+	private madeAttachementField(humans: Human[]): Human[] {
```
etokareva: can we create madeFiel method and pass fiel name to it?

PR #99, opened 2017-12-04
Update core to the latest

klumba12:


core/column/column.service.js
```diff
@@ -101,9 +102,24 @@ export function collapse(view) {
 
 export function widthFactory(table, form) {
 	const layout = table.model.layout;
+	const columns = table.data.columns();
+	const columnMap = map(columns);
 	form = form || layout().columns;
 
+	const occupied = columns
+		.filter(c => form.hasOwnProperty(c.key) || ('' + c.width).indexOf('%') < 0)
```
etokareva: can we check this inside of reduce? Looks like we don't need filtered array itself

core/column/column.service.js
```diff
@@ -113,22 +129,33 @@ export function widthFactory(table, form) {
...
-			width = rect.width * percent / 100;
+	function getWidth(column) {
+		let size = column;
+		if (form.hasOwnProperty(column.key)) {
+			size = form[column.key];
 		}
 
-		return Math.max(parseInt(width), parseInt(column.minWidth || 20)) + 'px';
+		let width = size.width;
+		if (width || width === 0) {
```
etokareva: width >= 0 ?

core/filter/filter.model.js
```diff
@@ -5,9 +5,15 @@ import {noop} from '../utility';
...
 		this.resource = new Resource();
+
 		this.by = {};
 		this.match = match;
 		this.fetch = noop;
 		this.unit = 'default';	// default|row
+		this.assertFactory = () => ({
+			equals: (x, y) => x === y,
+			lessThan: (x, y) => x < y,
+			isNull: x => x === '' || x === null || x === undefined || isNaN(x) ||  isFinite(x)
```
etokareva: is it correct name for such function?

core/infrastructure/composite.js
```diff
@@ -0,0 +1,36 @@
...
+
+	static command(list) {
+		return new Command({
+			source: 'composite',			
+			canExecute: (...args) => {
+				return list.reduce((memo, cmd) => memo || cmd.canExecute(...args), false);
+			},
+			execute: (...args) => {
+				return list
+					.filter(cmd => cmd.canExecute(...args))
```
etokareva: we could use only reduce there

core/services/path.js
```diff
@@ -1,13 +1,9 @@
...
 	const last = parts.length - 1;
-	const accessor =
-		parts.filter((part, index) => index !== last)
-			.reduce((accessor, part) =>
-					accessor
-						? entry => accessor(entry)[part]
-						: entry => entry[part],
-				null);
+	const accessor = parts
+		.filter((part, index) => index !== last)
```
etokareva: use this check inside of reduce?

PR #100, opened 2017-12-06
Update view ctrl and column list

klumba12:


src/infrastructure/design/singleton.ts
```diff
@@ -0,0 +1,26 @@
+import { AppError } from 'ng2-qgrid/core/infrastructure';
+
+function decorate(method) {
+	let instance;
+	let hasInstance = false;
+	return function value() {
```
etokareva: use arrow function?

klumba12: i can't do it, cause this is used inside)

PR #102, opened 2017-12-18
Style fixes for column chooser

klumba12:


.jsbeautifyrc
```diff
@@ -0,0 +1,7 @@
+{
```
vkorolev: Isn't editor config enough?

PR #103, opened 2017-12-18
Style fixes for column sort and column filter

klumba12:


plugin/column-sort/column.sort.view.js
```diff
@@ -1,5 +1,5 @@
 import {PluginView} from '../plugin.view';
-import {EventListener, EventManager} from '../../core/infrastructure';
+import {EventListener,EventManager} from '../../core/infrastructure';
```
etokareva: I would like to have one formatting style in the project 😈

PR #105, opened 2017-12-19
Column templates application improvements

klumba12:


core/scene/scene.model.d.ts
```diff
@@ -4,6 +4,7 @@ import {RowView} from './view/row.view';
 export declare class SceneModel {
 	constructor();
 
+    round: number;
 	status: string;
```
vkorolev: Doesn't seem to be formatted

core/services/path.js
```diff
@@ -1,13 +1,10 @@
 export function compile(path) {
 	const parts = path.split('.');
-	const last = parts.length - 1;
-	const accessor = parts
-		.filter((part, index) => index !== last)
-		.reduce((accessor, part) => accessor ? entry => accessor(entry)[part] : entry => entry[part], null);
-
+    const last = parts.length - 1;
+    const accessor = getAccessor(parts, last);
```
vkorolev: Different formats here too

PR #106, opened 2017-12-20
Filter row

klumba12:


core/head/head.view.js
```diff
@@ -70,12 +70,11 @@ export class HeadView extends View {
...
 			source: 'head.view',			
 			canExecute: () => true,
-			execute: e => {
-				const key = e.source.sourceKey;
+			execute: (column, search) => {
 				const filter = this.model.filter;
-				const by = clone(filter().by);
-				const search = e.search;
-				if (search.length) {
+                const by = clone(filter().by);
```
titov-max: excess tab before ```const```

core/column-type/filter.row.column.js
```diff
@@ -20,8 +20,7 @@ export class FilterRowColumnModel extends ColumnModel {
 		this.class = 'control';
 
 		this.canFilter = model.canFilter && model.class === 'data';
-		this.sourceKey = model.key;
-		this.sourceType = model.type;
+		this.model = model;;
```
titov-max: excess ; in the end

PR #116, opened 2017-12-25
Fix a bug in array edit form layout

Windman:


src/themes/material/templates/edit-cell-array.tpl.html
```diff
@@ -6,20 +6,20 @@
...
-                                  selectable="true"
-                                  removable="true"
-                                  (remove)="$view.edit.cell.value.splice($index, 1)">
-                            {{value}}
-                            <mat-icon class="q-grid-icon"
-                                      matChipRemove>cancel</mat-icon>
-                        </mat-chip>
-                    </mat-chip-list>
+                <mat-chip-list #chipList>
+                    <mat-chip class="q-grid-array-view__chiplist_chip"
```
klumba12: rename class to `q-grid-array-item`

PR #117, opened 2017-12-25
legend plugin implemented

munkk:


src/themes/material/assets/legend.scss
```diff
@@ -28,4 +28,17 @@ ul.q-grid-legend-list {
 	height: 10px;
 	border-radius: 10px;
 	margin: auto;
-}
\ No newline at end of file
+}
+
+.female {
```
klumba12: remove male femle undefined classes from here, this is just an example

src/themes/material/templates/plugin-legend.tpl.html
```diff
@@ -0,0 +1,16 @@
+<ng-template key="plugin-legend.tpl.html">
+	<ul class="q-grid-legend-list">
```
klumba12: this should be empty by default

src/themes/material/templates/toolbar-bottom.tpl.html
```diff
@@ -1,3 +1,10 @@
 <ng-template key="bottom-toolbar.tpl.html">
-	<q-grid-pager></q-grid-pager>
+	<div class="q-grid-toolbar">
```
klumba12: please make layout like here: https://github.com/qgrid/ng/blob/master/src/themes/material/templates/toolbar.bottom.html

it will be easier to support

plugin/legend/legend.model.js
```diff
@@ -0,0 +1,7 @@
+import {Resource} from '../../core/resource';
```
klumba12: there is no need in plugin/legend, it doesnt contain any addition logic, please remove all this 3 files

src/plugins/legend/legend.component.ts
```diff
@@ -0,0 +1,24 @@
...
+export class LegendComponent extends PluginComponent implements OnInit, OnDestroy {
+
+	constructor(@Optional() root: RootService) {
+		super(root);
+
+		this.models = ['legend'];
+	}
+
+	ngOnInit() {
+		this.context = {$implicit: this};
```
klumba12: this is by default in PluginComponent :)

src/plugins/legend/legend.component.ts
```diff
@@ -0,0 +1,24 @@
...
+	}
+
+	ngOnDestroy() {
+		super.ngOnDestroy();
```
klumba12: no need to override

PR #120, opened 2017-12-25
Right click and mouse wheel click on backdrop now close editor

munkk:


plugin/backdrop/backdrop.view.js
```diff
@@ -1,25 +1,33 @@
 import {PluginView} from '../plugin.view';
 import {Event, EventListener, EventManager} from '../../core/infrastructure';
 
+const mouse_left_button = 1;
```
klumba12: Make it `LEFT_BUTTON`, `WHEEL_BUTTON` in upper case

PR #123, opened 2017-12-25
Apply theme dynamically

klumba12:


src/themes/material/theme.module.ts
```diff
@@ -52,11 +46,16 @@ import { PipeModule } from 'ng2-qgrid/pipes';
...
+	providers: [ThemeService],
+	entryComponents: [ThemeComponent]
 })
 export class ThemeModule {
-	constructor() {
+	constructor(
+		theme: ThemeService,
+		componentResolver: ComponentFactoryResolver
+	) {
+		theme.name = 'material';
```
vkorolev: Probably, it's a responsibility of client code, because if there are multiple themes required, the last one is used.

vkorolev:
Does it work properly after build:aot?

PR #135, opened 2018-01-11
Add dark color theme

Windman:


src/theme/material/assets/dark-colors.scss
```diff
@@ -0,0 +1,8 @@
+$color-less: rgba(0, 0, 0, 0);
+
+$dark: #424242;//rgba(0, 0, 0, 0.87);
+$dark-accent: #616161;
+$dark-text: #fff;
+//$dark-default: rgba(255, 255, 255, 0.5);
+$dark-header:  hsla(0,0%,100%,.7);
```
etokareva: looks like this is the same as rgba(255, 255, 255, 0.7). We use rgb, rgba for opacity and hex values. Hsla is interesting way to define colors, but it may confuse a bit

PR #136, opened 2018-01-12
Add animation for a sort direction

Windman:


plugin/column-sort/column.sort.view.js
```diff
@@ -22,12 +22,30 @@ export class ColumnSortView extends PluginView {
 				this.clear();
```
klumba12: use tabs instead of spaces

plugin/column-sort/column.sort.view.js
```diff
@@ -22,12 +22,30 @@ export class ColumnSortView extends PluginView {
...
 
 				const view = this.view;
-				const display = this.display;
-				if (view.sort.order(this.column) >= 0) {
+                const display = this.display;
+                const iconDesc = context.iconDesc;
+                const iconAsc = context.iconAsc;
+                const styleName = `icon-visible`;
```
klumba12: use q-grid prefix

klumba12: consider to rename to `q-grid-active`

klumba12: `styleName `rename to activeClass`

PR #141, opened 2018-01-19
Add new sort format

Windman:


core/sort/sort.model.js
```diff
@@ -7,4 +8,38 @@ export class SortModel {
 		this.mode = 'multiple';
 		this.trigger = ['reorder'];
 	}
+
+	tryToResolveSortFormat() {
```
klumba12: i think this is unnecessary to give our user such a possibility to rewrite also, sort.service is more common staff, can you change key, value functions in sort.service instead of sort.model + sort.view?
also if you change sort.by property new invalidate is called that is not good

core/sort/sort.model.js
```diff
@@ -1,4 +1,5 @@
 import {Resource} from '../resource';
+import { isString, isObject } from '../utility';
```
klumba12: remove this line

core/services/pair2.js
```diff
@@ -0,0 +1,57 @@
...
+	return pairs.map(key).findIndex(k => k === pairKey);
+}
+
+export function value(pair) {
+	let value = "";
+	if (isObject(pair) && !isString(pair)) {
+		const pairKey = key(pair);
+		value = pair[pairKey];
+	} else if (isString(pair)) {
+		let delimiterSet = [{'desc':'-'}, {'asc':'+'}];
```
klumba12: use const everywhere if variable is not changed

.gitignore
```diff
@@ -81,3 +81,4 @@ package-lock.json
 
 # Generated HTML #
 *.gen.html
+.vscode/settings.json
```
klumba12: it should be in git

core/sort/sort.service.js
```diff
@@ -22,4 +22,58 @@ export function orderFactory(model) {
...
+	let key = '';
+	if (!isString(pair)) {
+		key = Object.keys(pair)[0];
+	} else {
+		key = pair.split(/[+-]/)[1];
+	}
+	
+	if(!key){
+		throw new AppError(
+			'pair',
```
klumba12: sort.service

core/sort/sort.service.js
```diff
@@ -22,4 +22,58 @@ export function orderFactory(model) {
...
+}
+
+function newValue(pair) {
+	let value = '';
+	if (!isString(pair)) {
+		const pairKey = key(pair);
+		value = pair[pairKey];
+	} else {
+		let delimiterSet = [{'desc':'-'}, {'asc':'+'}];
+		let direction = pair.split(pair.split(/[+-]/)[1])[0];
```
klumba12: use const :)

core/sort/sort.service.js
```diff
@@ -22,4 +22,58 @@ export function orderFactory(model) {
...
+		throw new AppError(
+			'pair',
+			`Key is not defined in "${pair}"`);
+	}
+
+	return key;
+}
+
+function newValue(pair) {
+	let value = '';
```
klumba12: consider to use 2 returns instead of `value`, in this case it will separate logic better

core/sort/sort.service.js
```diff
@@ -22,4 +22,58 @@ export function orderFactory(model) {
...
+				}
+			}).reduce((p, k) => k, '');
+		}).filter(v => v)
+		.reduce((p, k) => k, '');
+	}
+	
+	return value;
+}
+
+function newMap(pairs) {
```
klumba12: newMap and newIndex are strange names

PR #142, opened 2018-01-22
phoenix custom theme

munkk:


demo/app/phoenix-theme.scss
```diff
@@ -0,0 +1,96 @@
+@import '~@angular/material/theming';
```
klumba12: put it under theme/assets and call it phoenix.scss

PR #162, opened 2018-02-26
D.ts interface to function

munkk:


core/filter/match.d.ts
```diff
@@ -1,2 +1 @@
-export declare type IMatch = (x: any, value: any) => boolean;
-
+export declare function match(context: any): (x: any, value: any) => boolean;
```
klumba12: remove this file, put declaration to filter.model.d.t.s

PR #166, opened 2018-03-21
fix numpad issue

munkk:


src/plugins/pagination/pager-target.component.ts
```diff
@@ -13,37 +13,32 @@ export class PagerTargetComponent extends PluginComponent {
 		super(root);
 	}
 
-	private value = '';
+	private value: number = this.current;
```
klumba12: this should be init in ngOnInit, how does it work?))

src/plugins/pagination/pager-target.component.ts
```diff
@@ -13,37 +13,32 @@ export class PagerTargetComponent extends PluginComponent {
...
+			case 'right':
+			case 'backspace': {
 				break;
 			}
 			default: {
 				const digit = Number.parseInt(code);
-				const total = this.total();
-				const page = Number.parseInt(this.value + digit);
+				const total = this.total;
+				const page = value + '' + digit || 0;
```
klumba12: '' + value + digit, why we need to add here || 0? value shoudn't be empty ever

src/common/focus/focus.directive.ts
```diff
@@ -31,6 +31,6 @@ export class FocusDirective implements AfterViewInit {
 
 		// we need a small timeout to wait, for example, position directive
 		// in other case it will scroll to element before layout
-		this.zone.runOutsideAngular(() => setTimeout(() => element.focus(), 10));
+		this.zone.runOutsideAngular(() => setTimeout(() => element.focus(), 100));
```
klumba12: revert this changes

PR #175, opened 2018-03-29
add focus to after filter/column-chooser events

munkk:


src/plugins/colum-chooser/column-chooser.component.ts
```diff
@@ -20,6 +20,12 @@ export class ColumnChooserComponent extends PluginComponent implements OnInit, O
 		super(root);
 
 		this.models = ['columnChooser'];
+
+		this.using(this.model.sceneChanged.on(e => {
```
klumba12: ok, so, can we have a separate FocusService that will do next for now:
1. Injectable to components
2. have a method `focusService.activateAfterRender`

src/plugins/focus.service.ts
```diff
@@ -0,0 +1,41 @@
...
+		this.model = root.model;
+	}
+
+	private model;
+
+	private focus() {
+		this.root.table.view.focus();
+	}
+
+	activateAfterRender(state: string, ...args) {
```
titovmx: More common to place public methods before private. It allows to find methods which can be used to serve programmer purposes easier.

munkk: ok)

PR #186, opened 2018-04-13
Add rest plugin

vkorolev:


plugin/rest/rest.model.js
```diff
@@ -0,0 +1,13 @@
+import {Resource} from '../../core/resource';
+import {identity, cloneDeep, isArray, isObject, isDate, isBoolean, isNumber, isFunction} from '../../core/utility';
+
+export class RestModel {
+	constructor() {
+		this.resource = new Resource();
+
+		this.url = '';
+		this.method = 'get';
+		this.serialize = null;
```
klumba12: can we have serialize set here?

src/plugins/rest/rest.component.ts
```diff
@@ -0,0 +1,37 @@
...
+		super(root);
+
+		this.models = ['rest'];
+	}
+
+	ngOnInit() {
+		super.ngOnInit();
+
+		const {method, url, serialize} = this.model.rest();
+		this.rest = new RestView(this.model, {
```
klumba12: no need to pass url, method, serialize, it should be inside the model automatically

plugin/rest/rest.js
```diff
@@ -1,52 +1,49 @@
...
 
 export class Rest extends PluginView {
-	constructor(model, config) {
+	constructor(model, {get, post}) {
 		super(model);
 
-		const settings = merge({
-			method: 'get'
-		}, config);
+		const {method, url} = this.model.rest();
```
klumba12: can put serialize here?

plugin/rest/rest.js
```diff
@@ -1,52 +1,49 @@
...
+		const fetch = this.fetchFactory(method, get, post);
+		const serialize = this.serializeFactory(method, model.rest().serialize);
```
klumba12: and call it doSerialize :)

plugin/rest/rest.js
```diff
@@ -1,52 +1,49 @@
...
 		});
 	}
 
-	fetchFactory(settings) {
-		if (settings.method.toLowerCase() === 'get') {
-			return settings.get;
-		} else if (settings.method.toLowerCase() === 'post') {
-			return settings.post;
+	fetchFactory(method, get, post) {
+		if (method.toLowerCase() === 'get') {
```
klumba12: it's ideal to use switch, case here...

src/plugins/rest/rest.component.ts
klumba12: remove `url, method, serialize` from here

PR #198, opened 2018-04-17
QB toolbar

klumba12:


src/plugins/expression-builder/eb-node.component.ts
```diff
@@ -1,10 +1,53 @@
...
 	selector: 'q-grid-eb-node',
 	templateUrl: './eb-node.component.html'
 })
-export class EbNodeComponent {
-	@Input() public model: Node;
+export class EbNodeComponent implements OnInit, OnDestroy {
+	@Input() model: Node;
+	@Input() parent: EbNodeComponent;
+
+	self = this;
```
vkorolev: i don't think, `self` is necessary here

src/plugins/expression-builder/eb-node.component.ts
```diff
@@ -1,10 +1,53 @@
...
+	element: HTMLElement;
+	children = new Array<EbNodeComponent>();
```
vkorolev: consider to move initialization to constructor and simplify it: `this.children = [];`

src/plugins/expression-builder/eb-node.component.ts
klumba12: it's used in template

src/plugins/expression-builder/eb-node.component.ts
klumba12: what is point to make like this? if you dont like it i can do `children: Array<EBNodeComponent> = []` :)

PR #208, opened 2018-04-19
Batch

munkk:


src/main/core/cell/cell-handler.component.ts
```diff
@@ -120,7 +122,7 @@ export class CellHandlerComponent implements OnInit, AfterViewInit {
 	}
 
 	startBatchEdit(e) {
-		this.root.table.view.addClass('q-grid-noselect');
+		this.root.model.selection({mode: 'range'});
```
klumba12: you need to persist selectionMode every time on batchEdit,
the same for editMode

core/body/body.ctrl.js
```diff
@@ -132,12 +132,12 @@ export class BodyCtrl extends View {
 				this.navigate(cell);
 				if (cell.column.editorOptions.trigger === 'click' && this.view.edit.cell.enter.canExecute(cell)) {
 
-					if (edit().method === 'batch' && this.selection.items.length > 1) {
-						this.model.selection({items: []});
+					if (edit().startCell) {
+						edit({startCell: null});
```
klumba12: you should memorize startCell, in cell-handler.component, you can get it there defenetly :)

PR #215, opened 2018-04-23
fix batch

munkk:


src/main/core/cell/cell-handler.component.ts
```diff
@@ -111,7 +111,7 @@ export class CellHandlerComponent implements OnInit, AfterViewInit {
 
 				if (model.edit().method === 'batch') {
 					if (prevCell) {
-						Fastdom.mutate(() => prevCell.removeChild(this.marker.nativeElement));
+						Fastdom.mutate(() => this.marker.nativeElement.parentNode.removeChild(this.marker.nativeElement));
```
klumba12: this.marker.nativeElement.remove()

src/main/core/cell/cell-handler.component.ts
```diff
@@ -1,4 +1,4 @@
-import { AfterViewInit, Component, ElementRef, HostListener, OnInit, ViewChild } from '@angular/core';
+import {AfterViewInit, Component, DoCheck, ElementRef, HostListener, OnInit, ViewChild} from '@angular/core';
```
klumba12: for formatting, where is docheck is used?

PR #223, opened 2018-05-14
Column filter fix

munkk:


src/plugins/column-filter/column-filter-by.component.ts
```diff
@@ -10,12 +10,23 @@ import { ColumnModel } from 'ng2-qgrid/core/column-type/column.model';
...
 		super(root);
 	}
 
+	get isBlanks() {
+		return this.byBlanks;
+	}
+
 	remove(item: string): void {
-		this.by.delete(item);
+		if (item) {
```
klumba12: make for byBlanks another method removeByBlanks, because in theory item can be an empty string

PR #227, opened 2018-05-17
Batch fix

munkk:


src/core/edit/edit.service.js
```diff
@@ -3,43 +3,46 @@ import {EditCellView} from './edit.cell.view';
 import {CommandManager} from '../command/command.manager';
```
klumba12: fox formatting

src/core/edit/edit.service.js
```diff
@@ -3,43 +3,46 @@ import {EditCellView} from './edit.cell.view';
...
+    const value = startCell.value;
 
-				const cell = cellView.model;
-				const type = cell.column.type;
-				if (startColumnType === type) {
-					const editor = new CellEditor(cell);
-					editor.label = label;
-					editor.value = value;
-					editView.editor = editor;
+    const rows = view.rows;
```
klumba12: const {rows, columns} = view;

src/core/edit/edit.service.js
```diff
@@ -3,43 +3,46 @@ import {EditCellView} from './edit.cell.view';
...
-						editView.batchCommit.execute();
-					}
-				}
-			}
-		}
-		finally {
-			editView.dispose();
-		}
-	}
+      for (let i = 0, max = selectionItems.length; i < max; i++) {
```
klumba12: reaname max to length

src/lib/main/core/cell/cell-handler.component.ts
```diff
@@ -7,148 +7,151 @@ import { RootService } from '../../../infrastructure/component/root.service';
 import { ViewCoreService } from '../../../main/core/view/view-core.service';
```
klumba12: fix formatting

src/lib/main/core/cell/cell-handler.component.ts
```diff
@@ -3,20 +3,22 @@ import { jobLine } from 'ng2-qgrid/core/services/job.line';
...
 })
 export class CellHandlerComponent implements OnInit, AfterViewInit {
 	private job = jobLine(150);
+	private markerJob = jobLine(150);
 	private startCell: CellView = null;
 	private initialSelectionMode: string = null;
 	private initialEditState: string = null;
 
 	constructor(private element: ElementRef, private root: RootService, private view: ViewCoreService) {
+		element.nativeElement.style.visibility = 'hidden';
```
klumba12: use fastdom.mutate for dom operations

src/core/services/value.js
```diff
@@ -19,7 +19,9 @@ export function getFactory(column) {
 			? row => column.value(row)
 			: column.path
 				? compileGet(column.path)
-				: row => row[column.key];
+				: row => row.rows
```
klumba12: this is wrong, what if user will pass object with property `rows`? you will give wrong value in this case, also why you need to make it right here?

PR #229, opened 2018-05-17
Clipboard

munkk:


src/core/row/row.selector.js
```diff
@@ -0,0 +1,208 @@
...
+        const x = specificTitles.indexOf(label);
+
+        body[y][x] = label;
+      }
+    }
+
+    return body;
+  }
+
+  itemsTitles(items) {
```
klumba12: please rename it

src/core/clipboard/clipboard.service.js
```diff
@@ -0,0 +1,96 @@
...
+export class ClipboardService {
+  static copy(selector) {
+    const table = buildTable(selector);
+    select(table);
+    table.remove();
+  }
+}
+
+function buildTable(selector) {
+  const rows = selector.rows.body;
```
klumba12: const {...} = selector.rows;

src/core/clipboard/clipboard.service.js
```diff
@@ -0,0 +1,96 @@
...
+  const head = selector.rows.head;
+  const foot = selector.rows.foot;
+  const source = selector.source;
+  let table = document.createElement('table');
```
klumba12: make const

PR #235, opened 2018-05-29
Proxy + bug fixes

klumba12:


src/core/column/column.factory.js
```diff
@@ -1,29 +1,29 @@
 import { Guard } from '../infrastructure/guard';
 import { assignWith, clone, isUndefined } from '../utility/kit';
 import { ColumnView as CustomColumn } from '../scene/view/column.view';
-import { TextColumn } from '../column-type/text.column';
-import { NumberColumn } from '../column-type/number.column';
+import { ArrayColumn } from '../column-type/array.column';
```
etokareva: Can we create barrel file inside of column-type folder?

src/core/view/view.factory.js
```diff
@@ -1,24 +1,24 @@
+import { Model } from '../infrastructure/model';
+import { ModelProxy } from '../infrastructure/model.proxy';
```
etokareva: and barrels inside of folders to reduce rows with import

PR #242, opened 2018-05-31
ExampleSelectMixBasicComponent component implemented

azkurban:


src/examples/select-mix-basic/example-select-mix-basic.component.html
```diff
@@ -0,0 +1,8 @@
+<q-grid header="Select Column Basic"
+        [rows]="rows | async"
+        selectionUnit="mix"
+        selectionMode="multiple"
+        selectionArea="body">
```
klumba12: remove please selectionArea it is `body` by default, just was showed in cell-basic to let user know about it

azkurban: fixed

PR #256, opened 2018-06-06
Column chooser cohort

munkk:


src/plugin/column-chooser/column.chooser.view.js
```diff
@@ -22,15 +22,37 @@ export class ColumnChooserView {
...
-
+		
+		this.toggleChild = new Command({
+			source: 'column.chooser',
+			execute: (parent, child) => {
+				const state = !this.state(child);
+				
+				child.isVisible = state;
+				
+				if (parent.children.some(c => this.state(c))) {
```
titovmx: Can you replace these two conditions by one?
if ```some``` will not find anything it returns false.

PR #265, opened 2018-06-09
Editors

klumba12:


src/core/format/format.service.js
```diff
@@ -6,4 +6,8 @@ export class FormatService {
     static date(x, format) {
         return x;
     }
+
+    static currency(X, format) {
```
titovmx: parameter shouldn't be capitalized

PR #283, opened 2018-06-14
Hierarchy Browser Example

munkk:


src/core/node/node.d.ts
```diff
@@ -42,4 +42,6 @@ export declare class Node {
 	 * Column key of node.
 	 */
 	source: string;
+
+	isVisited: boolean;
```
klumba12: remove this one

PR #288, opened 2018-06-18
Hierarchy example improvement

munkk:


src/core/column-type/group.column.js
```diff
@@ -24,7 +24,7 @@ export class GroupColumnModel extends ColumnModel {
 		this.path = 'key';
 		this.labelPath = 'key';
 		this.title = 'Group';
-		this.offset = 24;
+		this.offset = 40;
```
klumba12: revert this

src/theme/material/assets/node.scss
```diff
@@ -8,21 +8,18 @@ th.q-grid-group {
 
 td.q-grid-group,
 td.q-grid-row-expand {
```
klumba12: you need to do it in example scss not here

PR #290, opened 2018-06-20
ColumnTime_Fix

munkk:


src/core/edit/edit.cell.view.js
```diff
@@ -354,6 +354,22 @@ export class EditCellView {
 		return this.column.options;
 	}
 
+	set time(value) {
```
klumba12: you can't do it here :)

src/lib/common/time/time.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+    selector: '[q-grid-time]'
+})
+
+export class TimeDirective {
+    constructor(
+        private templateRef: TemplateRef<TimeContext>,
+        private viewContainerRef: ViewContainerRef
+    ) {
+        const context = new TimeContext();
+        viewContainerRef.createEmbeddedView(templateRef, context);
```
klumba12: createEmbeddedView(templateRef, this)

and put all context under this class

PR #301, opened 2018-06-22
Features/theme grid dark pr4

azkurban:
Dark theme: pager-target & cell-editors

### DIFF
--- a/src/examples/theme-grid-dark/example-theme-grid-dark.component.html
+++ b/src/examples/theme-grid-dark/example-theme-grid-dark.component.html
@@ -1,5 +1,7 @@
-<q-grid caption="Look People" class="q-grid-dark"
-        [rows]="rows | async">
+<q-grid caption="Look People" 
+	class="q-grid-dark"
+	[rows]="rows | async"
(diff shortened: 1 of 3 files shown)


src/core/edit/edit.cell.view.d.ts
```diff
@@ -23,6 +23,7 @@ export declare class EditCellView {
 
 	value: any;
 	label: any;
+	theme: string | Array<string>;
```
klumba12: it is not allowed to make it here

src/theme/material/templates/edit-cell-date.tpl.html
```diff
@@ -14,7 +14,7 @@
                 <mat-datepicker-toggle matSuffix
                                        [for]="picker">
                 </mat-datepicker-toggle>
-                <mat-datepicker #picker>
+                <mat-datepicker #picker [panelClass]="$view.edit.cell.theme">
```
klumba12: add a special service theme and call it instead of edit so it would be `$view.theme.class`

PR #309, opened 2018-06-25
autocomplete_plugin

munkk:


src/core/edit/edit.cell.editor.d.ts
```diff
@@ -10,10 +10,11 @@ export declare class CellEditor {
 	fetch: () => void;
 	resetFetch: () => void;
 
+	td: any;
```
klumba12: we can not propagate td here, it shoud be hidden

src/core/edit/edit.cell.view.d.ts
```diff
@@ -27,8 +27,9 @@ export declare class EditCellView {
 	readonly cell: CellView;
 	readonly row: any;
 	readonly column: ColumnModel;
-	readonly options: EditorOptions;
+	readonly options: EditorOptions;  
+	readonly editor: CellEditor;
 
-	readonly fetch: () => void;
+	readonly fetch: any;
```
klumba12: add all possible variants through the `|`

src/lib/plugins/autocomplete-editor/autocomplete-editor.component.ts
```diff
@@ -0,0 +1,73 @@
...
+
+		if (val === '') {
+			return;
+		}
+
+		return this.options.filter(option => option.toString().toLowerCase().includes(val.toLowerCase()));
+	}
+
+	private exit() {
+		this.cell.exit.execute(this.td);
```
klumba12: you can do just this.cell.exit.execute() without cell

src/lib/plugins/plugin.module.ts
```diff
@@ -52,7 +55,8 @@ Model.register('columnChooser', ColumnChooserModel)
 		RestModule,
 		StatusBarModule,
 		TabTrapModule,
-		TitleModule
+		TitleModule,
+		ReactiveFormsModule
```
klumba12: is it required by autocomplete? can you post a link

src/theme/material/templates/plugin-autocomplete-editor.tpl.html
```diff
@@ -0,0 +1,25 @@
+<ng-template key="plugin-autocomplete-editor.tpl.html"
+             let-$editor>
+   <q-grid-cell-editor class="q-grid-editor q-grid-position"
```
klumba12: cell editor itself should be in edt-cell-autocomplete tpl

src/core/edit/edit.cell.editor.d.ts
klumba12: use cell instead

src/lib/plugins/plugin.module.ts
munkk: its required by reactive forms, we could use template-driven forms instead, but with reactive forms it would be much easier to subscribe to value changes (as its said in docs https://material.angular.io/components/autocomplete/overview)

src/core/edit/edit.cell.view.d.ts
```diff
@@ -7,6 +7,7 @@ import { Model } from '../infrastructure/model';
 import { EditorOptions } from '../column-type/editor.options';
 import { KeyCode } from '../shortcut/key.code';
 import { CellView } from '../scene/view/cell.view';
+import { Fetch } from 'ng2-qgrid/core/infrastructure/fetch';
```
klumba12: ng2-qgrid is not allowed in the core

src/core/services/predicate.js
```diff
@@ -36,6 +36,9 @@ export function predicateFactory(search) {
 		}
 	}
 
-	const expr = new RegExp(search, 'gi');
-	return item => expr.test(item);
+	const pattern = escapeRegexp(search);
+	return item => {
+		const expr = new RegExp(pattern, 'gi');
```
klumba12: How it could affect performance?

src/lib/plugins/autocomplete-editor/autocomplete-editor.component.ts
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	filter(search: string) {
+		if (search === '') {
+			this.reset();
+			return;
+		}
+
+		const { type } = this.cell.column;
+		switch (type) {
```
klumba12: why we need this?

src/lib/plugins/autocomplete-editor/autocomplete-editor.component.ts
```diff
@@ -0,0 +1,85 @@
...
+			case 'number':
+			case 'text':
+			case 'date': {
+				const result = this.findItems(search);
```
klumba12: why result could be null or undef or...?

src/lib/plugins/autocomplete-editor/autocomplete-editor.component.ts
```diff
@@ -0,0 +1,85 @@
...
+					this.reset();
+				} else if (result.length) {
+					this.options = [result[0]];
+				}
+				break;
+			}
+		}
+	}
+
+	findItems(value) {
```
klumba12: put this body into the filter function

src/core/services/predicate.js
klumba12: Yes resetting lastIndex would be preferred way

src/lib/plugins/autocomplete-editor/autocomplete-editor.component.ts
klumba12: and what wrong with it?

src/lib/plugins/autocomplete-editor/autocomplete-editor.component.ts
```diff
@@ -0,0 +1,63 @@
...
+
+	filter(search: string) {
+		if (search === '') {
+			this.reset();
+			return;
+		}
+
+		const test = predicateFactory(search);
+		const result = this.items.filter(item => test(item));
+		if (result) {
```
klumba12: when it could be false? an empty array? but this is correct case

PR #310, opened 2018-06-25
Features/theme basic styling arrays

azkurban:


src/theme/basic/templates/body-cell-array-of-number.tpl.html
```diff
@@ -1,6 +1,6 @@
 <ng-template key='body-cell-array-of-number.tpl.html'
 			 let-$cell>
-	<ul>
+	<ul class="q-grid-array">
```
klumba12: remove q-grid-array class it's not necessary

azkurban: done in the last commit

PR #311, opened 2018-06-26
Example export

etokareva:


package.json
```diff
@@ -22,12 +22,14 @@
...
     "@angular/core": "^6.0.5",
-    "@angular/platform-browser": "^6.0.5",
-    "@angular/platform-browser-dynamic": "^6.0.5",
     "@angular/forms": "^6.0.5",
     "@angular/http": "^6.0.5",
+    "@angular/platform-browser": "^6.0.5",
+    "@angular/platform-browser-dynamic": "^6.0.5",
     "core-js": "^2.5.7",
+    "file-saver": "^1.3.8",
     "rxjs": "^6.2.1",
```
klumba12: add deps to devDependencies

src/core/plugin/plugin.service.js
```diff
@@ -1,4 +1,4 @@
-import {AppError} from '../infrastructure';
+import { AppError } from 'ng2-qgrid/core/infrastructure/error';
```
klumba12: in core it's restricted to use `ng2-qgrid` because core can be used in ng-qgrid too
use relative path

src/plugin/export/export.view.d.ts
```diff
@@ -0,0 +1,21 @@
+import { Model } from 'ng2-qgrid/core/infrastructure/model';
+import { Command } from 'ng2-qgrid/core/command/command';
+import { Row } from 'ng2-qgrid/core/dom/row';
+import { Column } from 'ng2-qgrid/core/dom/column';
+
+export declare class ExportView {
+	constructor(model: Model, context: { type: string });
+
+	model: {
+		[ key: string ]: any
```
klumba12: fix formatting to one line

PR #312, opened 2018-06-26
File upload fix

munkk:


src/lib/common/file/file-upload.directive.ts
```diff
@@ -54,11 +56,21 @@ export class FileUploadDirective extends NgComponent implements AfterViewInit {
...
 	}
 
+	onFocus(e) {
+		if (this.backdrop.hidden) {
+			this.revealBackdrop();
+		}
+	}
+
 	onClick() {
+		this.backdrop = document.querySelector('.q-grid-backdrop');
```
klumba12: what if there won't be a backdrop? e.g. user rewrites cell template and doesn't use backdrop?

PR #318, opened 2018-06-30
Export json bug

etokareva:


package.json
```diff
@@ -52,7 +52,7 @@
     "file-saver": "^1.3.8",
     "jasmine-core": "~2.99.1",
     "jasmine-spec-reporter": "~4.2.1",
-    "karma": "~1.7.1",
+    "karma": "^2.0.4",
```
klumba12: revert this back, cli works bad with new version

src/core/export/json/json.js
```diff
@@ -1,30 +1,18 @@
...
+			const flatRow = flatView(row);
+			const obj = {};
+			for (let column of columns) {
+				obj[column.title] = flatRow[column.key];
+			}
+			result.push(obj);
 		}
 
-		return updateTitles(result, titles);
+		return JSON.stringify(result, '', 4);
```
klumba12: make 2 as default indent

PR #323, opened 2018-07-03
Features/theme basic styling sorting

azkurban:
Basic Theme: sorting support added


src/theme/basic/templates/plugin-column-sort.tpl.html
```diff
@@ -1,5 +1,6 @@
 <ng-template key="plugin-column-sort.tpl.html" let-$columnSort>
-    <span class="q-grid-icon q-grid-asc">&uarr;</span>
-    <span class="q-grid-icon q-grid-desc">&darr;</span>
+    <i class="material-icons q-grid-icon q-grid-asc">arrow_upward</i>
```
klumba12: need to you UTF symbols no material is allowed in this theme

azkurban: fixed in the last commit

PR #328, opened 2018-07-04
Add basic support of edit row fucntionality

Windman:


src/theme/material/templates/edit-form-text.tpl.html
```diff
@@ -4,8 +4,8 @@
         <input matInput
                #input
                q-grid-focus
-               maxlength="{{$control.cellEditor.cell.column.maxLength}}"
+               maxlength="{{$control.cellEditor.td.column.maxLength}}"
```
klumba12: td is a private property it's not allowed to use it

PR #334, opened 2018-07-11
update edit model typings

titovmx:


src/core/edit/edit.model.d.ts
```diff
@@ -56,9 +56,21 @@ export declare interface EditModel {
 	 */
 	clear?: Command;
 
+	/**
+	 * Object that contains `{columnKey: keyboardKeys}` map, that is used by q-grid to manage
+	 * when cancel command should be execute on key down event.
+	 */
+	cancelShortcuts?: CommandShortcuts;
```
klumba12: make it like {[key:string]: string}

titovmx: Do you mean without separate type?

PR #337, opened 2018-07-14
Features/theme basic simple pager bugs fix

azkurban:
Theme Basic bug fix: pager changes its width when page size is changed


azkurban:
Is closed due to its changes will be included into next pull request

PR #339, opened 2018-07-15
Features/theme basic styling

azkurban:
Includes changes from previous pull request (#338)


src/examples/aggregate-column-basic/example-aggregate-column-basic.component.html
```diff
@@ -1,4 +1,4 @@
-<q-grid caption="Aggregate Column Basic"
+<q-grid class="q-grid-aggregation" caption="Aggregate Column Basic"
```
klumba12: get rid of this class

src/theme/basic/index.scss
```diff
@@ -13,9 +14,33 @@ $pager-primary: #757575;
 $icon-button-disabled: rgba(0,0,0,.26);
 $hidden: transparent;
 
+.q-grid-aggregation {
```
klumba12: you don't need to use specific class just apply for any tfoot, but not for floating rows

src/examples/aggregate-column-basic/example-aggregate-column-basic.component.html
azkurban: fixed in my last commits

src/theme/basic/templates/body-cell-select.tpl.html
```diff
@@ -0,0 +1,10 @@
+<ng-template key='body-cell-select.tpl.html' let-$cell let-$view="$view">
+	    <div>
```
klumba12: do we need div here?

src/theme/basic/index.scss
```diff
@@ -4,6 +4,7 @@ $head-sorted: #004559;
 
 $grid-border: #e0e0e0;
 $grid-array-background: #e0e0e0;
+$total-row-border: #e0e0e0;
```
klumba12: lets make similar name convention, for all variables?

PR #350, opened 2018-08-03
Validation

etokareva:


src/lib/plugins/validation/rule.component.ts
```diff
@@ -0,0 +1,78 @@
+import { Component, ContentChild, Input, OnChanges, SimpleChanges, TemplateRef } from '@angular/core';
+import { PluginService } from '../plugin.service';
+import { TemplateHostService } from '../../template/template-host.service';
+
+const ruleBindings = [ 'required', 'not_empty_list', 'any_object', 'eq', 'string', 'length_between', 'length_equal', 'min_length',
```
klumba12: we have only for and key are not rules? maybe easier will be to add them as excpetions in ngChanges rather than support this array?

src/lib/plugins/validation/validation.component.ts
```diff
@@ -0,0 +1,19 @@
...
+	template: '',
+	providers: [ PluginService ]
+})
+export class ValidationComponent {
+	@Input() type: string;
+
+	constructor(private plugin: PluginService) {
+	}
+
+	get resource() {
```
klumba12: remove this one, it's not used in by ng2

src/theme/material/templates/edit-cell-array.tpl.html
```diff
@@ -31,6 +31,12 @@
                        q-grid-focus
                        [matChipInputFor]="chipList" />
             </mat-form-field>
+	        <q-grid-validator
```
klumba12: something wrong with formatting

src/lib/plugins/validation/validator.component.ts
```diff
@@ -0,0 +1,29 @@
...
+@Component({
+	selector: 'q-grid-validator',
+	templateUrl: './validator.component.html',
+	providers: [ TemplateHostService, PluginService ]
+})
+export class ValidatorComponent implements OnInit {
+	@Input() value: string;
+	@Input() key: string;
+	@Input() type: string;
+	context = {};
```
klumba12: context: {$implicit: ValidatorView}

PR #353, opened 2018-08-14
Fix date editor

munkk:


src/lib/common/date/date.directive.ts
```diff
@@ -0,0 +1,26 @@
...
+@Directive({
+    selector: '[q-grid-date]'
+})
+export class DateDirective {
+    constructor(templateRef: TemplateRef<DateDirective>, viewContainerRef: ViewContainerRef) {
+        viewContainerRef.createEmbeddedView(templateRef, this);
+    }
+
+    isValid(text) {
+        if (text.search(/^\d{1,2}[\/|\-|\.|_]\d{1,2}[\/|\-|\.|_]\d{4}$/g) !== 0) {
```
klumba12: check if it's compatible with different locale

src/lib/common/date/date.directive.ts
```diff
@@ -0,0 +1,23 @@
+import { Directive, TemplateRef, ViewContainerRef } from '@angular/core';
+
+const DATEPATTERN_1 = /^\d{1,2}[\/|\-|\.|_]\d{1,2}[\/|\-|\.|_]\d{4}$/g; // Month/Date/Year Date/Month/Year
```
klumba12: have you investigated the settings that could be applied? I mean LOCALE?

src/lib/common/date/date.directive.ts
```diff
@@ -0,0 +1,21 @@
...
+@Directive({
+    selector: '[q-grid-date]'
+})
+export class DateDirective {
+    constructor(templateRef: TemplateRef<DateDirective>, viewContainerRef: ViewContainerRef) {
+        viewContainerRef.createEmbeddedView(templateRef, this);
+    }
+
+    isValid(text) {
+        if (!(text.search(DATEPATTERN_1) === 0) && !(text.search(DATEPATTERN_2) === 0)) {
```
klumba12: please make it positive, also don't need to write 'else' as we use else

klumba12: Dont we use tabs for formatting?

PR #361, opened 2018-08-16
Cell editor impr

munkk:


src/theme/material/assets/cell.array.scss
```diff
@@ -25,6 +25,12 @@ td.q-grid-array {
 	min-height: 48px;
 	width: 340px;
 	height: auto;
+	.q-grid-editor-content {
+		width: 86%;
```
klumba12: why we use here percents? it's not clear why we use here exactly these values?

PR #372, opened 2018-09-11
File backdrop fix

munkk:


src/lib/common/file/file.directive.ts
```diff
@@ -64,4 +67,24 @@ export class FileDirective extends NgComponent {
 			this.value = e.target.result;
 		}
 	}
-}
+
+	onClick(e) {
```
klumba12: could we directly call hide, reveal instead of using onClick and onFocus

src/lib/plugins/backdrop/backdrop.component.ts
```diff
@@ -25,7 +25,9 @@ export class BackdropComponent {
 		$implicit: this
 	};
 
-	constructor(element: ElementRef) {
+	constructor(backdropService: BackdropService, element: ElementRef) {
+		backdropService.element = element;
```
klumba12: do we need reset element onDestroy?

src/lib/plugins/backdrop/backdrop.service.ts
```diff
@@ -0,0 +1,26 @@
+import { Injectable, Renderer2, ElementRef } from '@angular/core';
+import { Guard } from 'ng2-qgrid/core/infrastructure/guard';
+import { jobLine } from 'ng2-qgrid/core/services/job.line';
+
+@Injectable()
+export class BackdropService {
+	element: ElementRef;
+	isVisible = true;
+	job = jobLine(300);
```
klumba12: what is a purpose of job here?

src/lib/plugins/backdrop/backdrop.service.ts
```diff
@@ -0,0 +1,28 @@
...
+
+	reveal() {
+		Guard.notNull(this.element, 'element');
+		if (!this.isVisible) {
+			this.element.nativeElement.classList.remove('q-grid-backdrop-inactive');
+			this.isVisible = true;
+		}
+	}
+
+	reset() {
```
klumba12: could we directly do element = null and git rid of reset , because it's became not dual you set element using one way and unset with another

src/lib/plugins/backdrop/backdrop.service.ts
```diff
@@ -0,0 +1,28 @@
+import { Injectable, Renderer2, ElementRef } from '@angular/core';
+import { Guard } from 'ng2-qgrid/core/infrastructure/guard';
+
+@Injectable()
+export class BackdropService {
+	element: ElementRef;
+	isVisible = true;
```
klumba12: rename it to isActive please and make it private

src/lib/common/file/file.directive.ts
```diff
@@ -64,4 +67,18 @@ export class FileDirective extends NgComponent {
...
+
+	hideBackdrop() {
+		if (this.backdropService) {
+			this.backdropService.hide();
+		}
+	}
+
+	revealBackdrop() {
+		if (this.backdropService) {
+			if (!this.backdropService.isVisible) {
```
klumba12: you don't need to worry about that.

munkk: but we have to)
we should either check this state (!this.backdropService.isVisible) and place setTimeout here, in file.directive.ts or
we can get rid of this checking here, but place setTimeout in BackdropService file.
otherwise, it won't work correctly due to invalid calling sequence.
or i just do it wrong)

klumba12: ok, rename it to isActive please

PR #383, opened 2018-11-12
Qgrid 254: Delete a cell handler

yufilatov:


src/theme/material/assets/nohandler.scss
```diff
@@ -0,0 +1,115 @@
+$duration: 0.15s;
```
klumba12: can we put it under existing focus.scss?

src/theme/material/assets/nohandler.scss
```diff
@@ -0,0 +1,115 @@
...
+.q-grid-nohandler,
+.q-grid-lcc {
+    font-family: Roboto;
+    .q-grid {
+        .q-grid {
+            &-title {
```
klumba12: 80% of code is redundant, why we need here q-grid-lcc h2 etc?

src/examples/action-bar-on-enter/example-action-bar-on-enter.component.html
```diff
@@ -1,7 +1,7 @@
 <q-grid caption="Action Bar On Enter"
         [rows]="rows | async"
         [model]="gridModel"
-        class="q-grid-embed"
+        class="q-grid-nohandler"
```
klumba12: lets call it like `q-grid-cell-handler-hide` for now

src/theme/material/assets/nohandler.scss
yufilatov: If I change focuss.scss, it will affect the whole qgrid. As I understood, we need it only in Action Bar On Enter. Is it right?

PR #390, opened 2019-02-25
added screenshot-tests for first examples

player00:


e2e/features/column-array-basic.old.feature
```diff
@@ -0,0 +1,28 @@
...
+		When I click delete-icon
+		Then Value is deleted
+		When I press Enter
+		Then Changes are saved
+
+
+
+	Scenario: Custom Template column does work correctly - delete
+		Given I am on "column-array-basic"
+		When I click Custom Template data cell
```
klumba12: why not the same `I click cell "Strings"[0]` ?

player00: Old feature file was deleted

PR #399, opened 2019-04-02
Cross browser support

k-kutepov:
These fixes only for issues which I currently found. Most of them was appeared in Firefox.

IE 11 support still in progress, currently - it's just can't be launch for now because IE doesn't support classes and related code syntax features, it's need to investigate deeper.

### DIFF
--- a/src/core/body/body.ctrl.js
+++ b/src/core/body/body.ctrl.js
@@ -7,6 +7,7 @@ import { LEFT_BUTTON, checkButtonCode } from '../mouse/mouse.code';
 
 const VERTICAL_SCROLL_CLASS = `${GRID_PREFIX}-scroll-vertical`;
 const HORIZONTAL_SCROLL_CLASS = `${GRID_PREFIX}-scroll-horizontal`;
+const DEFAULT_DELTA_Y = 100;
 
 export class BodyCtrl {
 	constructor(model, view, table, bag) {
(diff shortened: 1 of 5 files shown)


src/lib/common/vscroll/vscroll.directive.ts
```diff
@@ -11,6 +11,7 @@ export class VscrollDirective {
 
 	constructor(private elementRef: ElementRef, zone: NgZone, renderer: Renderer2) {
 		zone.runOutsideAngular(() => {
+			elementRef.nativeElement.addEventListener('DOMMouseScroll', (event) => this.onMozWheel(event), { passive: false });
```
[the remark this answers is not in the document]
k-kutepov: You're right, please look the changes in further commit. 
I've just set unified default value for scroll height equal 100px (as in Chrome by default) which will be applied across different [delta modes](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode) and browsers.

src/lib/assets/table.scss
```diff
@@ -29,6 +29,8 @@
             display: block;
             overflow-y: scroll;
```
etokareva: I know that this is not part of your changes, but could we change 
```  scss
overflow-y
```
and
```  scss
overflow-x
```
to ```  scss
overflow
```
?

Comment (k-kutepov) [reply src/lib/assets/table.scss]: I believe we can) Thanks.

File: src/core/body/body.ctrl.js [hunk 3]
```diff
@@ -63,7 +63,9 @@ export class BodyCtrl {
 
 			Fastdom.measure(() => {
 				const lower = table.view.scrollHeight() - table.view.height();
-				const top = Math.min(lower, Math.max(upper, scroll().top + e.deltaY));
+				const defaultDeltaY = 100; // as default value in Chrome
```
klumba12: Make it as a constant on top of file
DEFAULT_DELTA_Y

PR #400, opened 2019-04-04
scroll to active menu item

pavpka:


src/app/app.component.ts
```diff
@@ -9,10 +9,24 @@ import { APP_ROUTES } from '../examples/example.module';
...
+	menuItems: QueryList<ElementRef>;
+
+	ngAfterViewInit() {
+		setTimeout(() => {
+			const activeItem = this.findActiveItem();
+			activeItem.nativeElement.scrollIntoView({ block: "center" });
+		}, 0);
+	}
+
+	findActiveItem = (): ElementRef => {
```
klumba12: make it private and just like a method
```typescript
privatefindActiveItem() {
}
```

src/app/app.component.ts
```diff
@@ -9,10 +9,24 @@ import { APP_ROUTES } from '../examples/example.module';
...
-export class AppComponent implements OnDestroy {
+export class AppComponent implements AfterViewInit, OnDestroy {
 	mobileQuery: MediaQueryList;
 	examples: Routes = APP_ROUTES;
 
+	@ViewChildren(RouterLinkActive, { read: ElementRef })
+	menuItems: QueryList<ElementRef>;
+
+	ngAfterViewInit() {
+		setTimeout(() => {
```
klumba12: Why we need timeout here? if still need please run it out of angular scope

```typescript
zone.runOutOfAngular...
```

PR #404, opened 2019-04-15
Special column filters

k-kutepov:
Added extra filter operators for “Number” and “Date” column filters. 
Full browser supporting (excluding IE 11).


src/core/filter/filter.model.js
```diff
@@ -15,5 +15,36 @@ export class FilterModel {
 			lessThan: (x, y) => x < y,
 			isNull: x => x === '' || x === null || x === undefined
 		});
+
+		this.operators = (column) => {
+			switch (column.type) {
+				case 'date': return [
```
klumba12: move return to a new line and add `{}`

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -48,10 +52,22 @@ export class ColumnFilterComponent implements OnInit {
 
 	ngOnInit() {
 		const { model } = this.plugin;
-		const key = this.column.key;
+		const { column } = this;
+		const { key } = column;
 		const context = { key };
 
 		const columnFilter = new ColumnFilterView(model, context);
+		this.filterOperators = model.filter().operators(column);
```
klumba12: can we make getter? get operators() { return model.filter().operators(this.column); }

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -48,10 +52,22 @@ export class ColumnFilterComponent implements OnInit {
...
+		this.filterOperator = this.filterOperators.length && this.filterOperators[0];
```
klumba12: just call it operator the class already contains word `filter`

klumba12: `this.operator = this.operators[0]` is enough

src/plugin/column-filter/column.filter.view.js
```diff
@@ -83,15 +84,43 @@ export class ColumnFilterView {
 
 			submit: new Command({
 				source: 'column.filter.view',
-				execute: () => {
+				execute: (operator = null, value = null) => {
```
klumba12: don't need to setup default values here

src/theme/material/templates/plugin-column-filter-between.tpl.html
```diff
@@ -0,0 +1,19 @@
+<ng-template key="plugin-column-filter-between.tpl.html"
```
klumba12: -default-between

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -140,4 +156,84 @@ export class ColumnFilterComponent implements OnInit {
...
+			}
+		}
+	}
+
+	get hasOperators() {
+		return this.filterOperators && this.filterOperators.length > 1;
+	}
+
+	beautifyOperatorName(op) {
+		const lcAll = op.replace(/([a-z])([A-Z])/g, '$1 $2').toLowerCase();
```
klumba12: please make a pipe for this.

src/theme/material/templates/plugin-column-filter-search.tpl.html
```diff
@@ -0,0 +1,61 @@
+<ng-template key="plugin-column-filter-search.tpl.html"
```
klumba12: -default-contains

src/theme/material/templates/plugin-column-filter.tpl.html
```diff
@@ -1,81 +1,49 @@
...
-					</mat-checkbox>
-				</li>
-			</ul>
-			<div [ngSwitch]="$columnFilter.isEmpty()"
-				 class="q-grid-column-filter-by">
-				<ng-container *ngSwitchCase="false">
-					<div class="q-grid-column-filter-by-blank">
-						NO SELECTED FILTERS
-					</div>
+			<div [ngClass]="{ 'q-grid-column-filter-with-operators' : $plugin.hasOperators }">
```
klumba12: please add class with selected operator e.g. `q-grid-column-filter-contains`

src/core/filter/filter.model.d.ts
```diff
@@ -70,4 +71,9 @@ export declare interface FilterModel {
 	 * * `isNull` should return true if value means null.
 	 */
 	assertFactory?: () => Assert;
+	
+	/**
+	 * Returns an array of filter operators available for a certain column.
+	 */
+	operators?: (column : ColumnModel) => string[];
```
klumba12: please rename it to `operatorFactory`

src/lib/pipes/beautify.pipe.ts
```diff
@@ -0,0 +1,13 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+    name: 'beautify'
```
klumba12: should have qGrid prefix, and can we do like that

`text | qGridText('fromCamelCase')` or something like that

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -26,10 +26,13 @@ import { PluginService } from '../plugin.service';
 export class ColumnFilterComponent implements OnInit {
 	@Input() column: ColumnModel;
 	@Input() search = '';
+	@Input() exprValue = [];
```
klumba12: can we put on .view level? where items are located? and rename to `value`

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -140,4 +157,70 @@ export class ColumnFilterComponent implements OnInit {
...
 		this.vscrollContext.container.reset();
 	}
+
+	clear() {
+		this.search = '';
+		this.exprValue = [];
+		this.context.$implicit.reset.execute();
+	}
+
+	get operatorTemplateKey() {
```
klumba12: make it as function, because it explicitly changed by component itsefl

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -140,4 +157,70 @@ export class ColumnFilterComponent implements OnInit {
...
+		const { column, operator } = this;
+		switch (operator) {
+			case 'contains': {
+				return 'plugin-column-filter-default-contains.tpl.html';
```
klumba12: please make plugin-column-filter- as a constant

src/lib/plugins/column-filter/column-filter.component.ts
klumba12: templateKeySelector(operator) {}

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -140,4 +157,70 @@ export class ColumnFilterComponent implements OnInit {
...
+			}
+			default: {
+				return column.type === 'date'
+					? 'plugin-column-filter-date.tpl.html'
+					: 'plugin-column-filter-default.tpl.html';
+			}
+		}
+	}
+
+	get operatorValue() {
```
klumba12: exprValue, also can we move it to the .view

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -140,4 +157,70 @@ export class ColumnFilterComponent implements OnInit {
...
+					: this.exprValue[0];
+			}
+		}
+	}
+
+	get hasOperators() {
+		return this.operators && this.operators.length > 1;
+	}
+
+	yyyymmdd(date, separator) {
```
klumba12: I think we have something for this, pliease look at date editors that we have in cells

src/core/expression/predicate.visitor.js
```diff
@@ -40,22 +40,24 @@ export class PredicateVisitor extends Visitor {
...
+				rt = getType(r[0]);
+				parse = parseFactory(rt);
 				r.forEach(x => map.add('' + x));
 			} else {
-				rCastAs = identity;
+				parse = identity;
 			}
 		} else {
-			rCastAs = castAsFactory(r);
+			rt = getType(r);
```
klumba12: make `rt` local const

src/core/expression/predicate.visitor.js
```diff
@@ -40,22 +40,24 @@ export class PredicateVisitor extends Visitor {
...
+			parse = parseFactory(rt);
 		}
 
 		const equals = assert.equals;
```
klumba12: don't we need to parse `r`?

src/core/expression/predicate.visitor.js
```diff
@@ -40,22 +40,24 @@ export class PredicateVisitor extends Visitor {
...
 		const isNull = assert.isNull;
 		const lessThan = assert.lessThan;
-		const lessThanOrEquals = (x, y) => equals(x, y) || lessThan(x, y);
+		const lessThanOrEquals = (x, y) => equals(parse(x), parse(y)) || lessThan(x, y);
```
klumba12: please make {equals, isNull, lessThan} = assert;

src/lib/pipes/text.pipe.ts
```diff
@@ -0,0 +1,20 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { AppError } from 'ng2-qgrid/core/infrastructure/error';
+
+@Pipe({
+    name: 'qGridText'
+})
+export class TextPipe implements PipeTransform {
+    transform(item: string, inputFormatType: string): string {
```
klumba12: just format and make the type as format: 'fromCamelCase'

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -140,4 +144,39 @@ export class ColumnFilterComponent implements OnInit {
...
 		this.vscrollContext.container.reset();
 	}
+
+	clear() {
+		this.search = '';
+		this.context.$implicit.reset.execute();
+	}
+
+	operatorTemplateKey(op) {
+		let tplName;
```
klumba12: call it just `key` pls

src/theme/material/templates/plugin-column-filter-date-between.tpl.html
```diff
@@ -0,0 +1,34 @@
...
+
+	<mat-form-field class="q-grid-column-filter-input half"
+	                *q-grid-date="let isValid='isValid'"
+	                [ngClass]="{'mat-form-field-invalid': !isValid(inputLeft.value)}">
+		<input q-grid-focus
+		       matInput
+		       #inputLeft
+		       placeholder="From"
+		       [matDatepicker]="leftPicker"
+		       [ngModel]="$columnFilter.expression.right[0] | qGridDate:'yyyy-MM-dd'"
```
klumba12: can we have here $columnFilter.value? not expression. expression should be encapsulated from UI. value can be string | number | date | array, why not?

src/core/expression/predicate.visitor.js
k-kutepov: I think it doesn't make sense, because we're just parsing `l` to type which `r` already has.

src/theme/material/templates/plugin-column-filter-date-between.tpl.html
k-kutepov: that's would be fine if value can be string | number | date. But array type (exactly in case of 'between' operator) makes me slightly confuse. 
I this case we should have some kind of getter/setter logic for `value` property to switch its type between array and primitive. Because in _default-between.tpl_ and _date-between.tpl_ UI we have two fields, the first should refer ngModel to `blahblah[0]` and the last should refer to `blahblah[1]`. And it will raised an error if `value` was previously defined as number or Date, for example.
I can implement both getter and setter for `$columnFilter.value`. 
Or we can have `$columnFilter.left` and `$columnFilter.right` properties instead (provides avoid using [0] and [1] indexes in UI, but also required some additional logic to combine it to expression before submit).

klumba12: I'm not sure by switching operator type just change value type that will be enough, it's okey that value for different expression can handle different types. If expression will require some object structure for the value what we are going to do? 

I'm worring about this because it'll go to public and no chance to change this, because it's our API.

src/lib/pipes/text.pipe.ts
```diff
@@ -0,0 +1,20 @@
...
+import { AppError } from 'ng2-qgrid/core/infrastructure/error';
+
+@Pipe({
+    name: 'qGridText'
+})
+export class TextPipe implements PipeTransform {
+    transform(item: string, format: 'fromCamelCase'): string {
+        switch (format) {
+            case 'fromCamelCase': {
+                const lcAll = item.replace(/([a-z])([A-Z])/g, '$1 $2').toLowerCase();
```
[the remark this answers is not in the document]
k-kutepov: Lower Case All & Upper Case First :)) It seems you hate all my variables

src/core/expression/predicate.visitor.js
```diff
@@ -40,22 +40,19 @@ export class PredicateVisitor extends Visitor {
...
-		} else {
-			rCastAs = castAsFactory(r);
 		}
 
-		const equals = assert.equals;
-		const isNull = assert.isNull;
-		const lessThan = assert.lessThan;
-		const lessThanOrEquals = (x, y) => equals(x, y) || lessThan(x, y);
+		const { equals, isNull, lessThan } = assert;
+		const lessThanOrEquals = (x, y) => equals(parse(x), parse(y)) || lessThan(x, y);
```
klumba12: why in one case we use parse and in another not?

src/core/expression/predicate.visitor.js
```diff
@@ -70,25 +67,25 @@ export class PredicateVisitor extends Visitor {
 				predicate = l => isNull(l);
 				break;
 			case 'equals':
-				predicate = l => equals(rCastAs(l), l);
+				predicate = l => equals(parse(l), r);
```
klumba12: i can't see code where we parse r? don't we need to do it?

src/core/expression/predicate.visitor.js
```diff
@@ -1,5 +1,5 @@
 import { AppError } from '../infrastructure/error';
```
klumba12: please verify that all tests are green

src/lib/pipes/text.pipe.ts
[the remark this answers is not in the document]
klumba12: just wasn't able to get it :) 

please remove first letter up, it can be done in css. 
and you will not need to use variables at all.

src/lib/plugins/column-filter/column-filter.component.ts
```diff
@@ -136,8 +140,59 @@ export class ColumnFilterComponent implements OnInit {
 		};
 	}
 
-	reset() {
+	reset(op) {
+		let { value } = this.context.$implicit;
```
klumba12: this is not good place to put this logic

klumba12: also it should be done in `.view.js`

src/core/expression/predicate.visitor.js
k-kutepov: Okay, you're absolutely right. I just was not able to realize why we should parse `r` to itself type. But `rt` technically is not equals to real `r` type and we really need to parse it everywhere. Thanks.

src/core/expression/predicate.visitor.js
k-kutepov: Indeed, it slightly weird step which obviously was bound with parsing of `r` ) Fixed in next commit. Thanks

src/core/expression/predicate.visitor.js
k-kutepov: Confirm that.

>  PredicateVisitor
>     visitGroup
>       √ check for equality with op = "and"
>       √ check for equality with op = "or"
>     visit condition
>       √ check for equality with op = "notEquals"
>       √ check for equality with op = "greaterThan"
>       √ check for equality with op = "lessThan"
>       √ check for equality with op = "between" inside range
>       √ check for equality with op = "between" outside range
>       √ check for equality with op = "between" outside range
>       √ check for equality with op = "between" on bounds of range
>       √ check for equality with op = "between" on bounds of range
>       √ check for equality with op = "in"
>       √ check whether "findSomeString" contains "Some"
>       √ "findSomeString" should not contains "text"
>       √ "findSomeString" should starts with "find"
>       √ "findSomeString" should ends with "String"
>       √ should return true if 101 greater than or equals 100
>       √ should return true if 99 less than or equals 100
>       √ should return value if not null
>       √ should return true if null

src/theme/material/templates/plugin-column-filter-date.tpl.html
```diff
@@ -0,0 +1,21 @@
...
+
+	<mat-form-field class="q-grid-column-filter-input"
+	                *q-grid-date="let isValid='isValid'"
+					[ngClass]="{'mat-form-field-invalid': !isValid(input.value)}">
+					
+		<input q-grid-focus
+		       matInput
+		       #input
+		       [matDatepicker]="picker"
+		       [placeholder]="$columnFilter.expression.op | qGridText : 'fromCamelCase'"
```
klumba12: can we mark expression as private and not use in template? it forces user to know expression structure during development

src/plugin/column-filter/column.filter.view.js
```diff
@@ -81,17 +88,46 @@ export class ColumnFilterView {
 				}
 			}),
 
+			update: new Command({
```
klumba12: can we call it changeOperator()

PR #406, opened 2019-04-16
add menu search

pavpka:


src/app/app.filter.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+@Pipe({
+    name: 'filterSearch'
+})
+
+export class FilterSearch implements PipeTransform {
+    transform(item: string, search: string) {
+        if (item && search) {
+            const contains = new RegExp(search, 'gi');
+            if (contains.test(item)) {
+                return true;
```
k-kutepov: Minor note: can we simplify this chunk with "return contains.test(item)" instead if-statement?

src/app/app.highlight.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+    transform(item: string, search: string) {
+        if (search && item) {
+            const contains = new RegExp(search, 'gi');
+            if (contains.test(item)) {
+                return item.replace(contains, elem => `<span class='highlight'>${elem}</span>`);
+            }
+        } else {
+            return item;
+        }
+    }
```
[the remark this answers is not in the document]
pavpka: It looks great. However the "search" value can still be empty, so what about add check of "search":
`return (contains.test(item) && search)
...` ?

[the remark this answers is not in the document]
k-kutepov: Indeed, you're right. I like this approach.

src/app/app.highlight.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
```
klumba12: please use highlight pipe it already exists

src/app/app.component.html
```diff
@@ -1,24 +1,31 @@
...
+				 [fixedInViewport]="mobileQuery.matches"
+				 fixedTopGap="56">
+		<mat-form-field class="search">
+			<input matInput
+				   #searchInput
+				   placeholder="Search"
+				   type="search" />
+		</mat-form-field>
 		<mat-nav-list>
 			<ng-container *ngFor="let example of examples">
```
klumba12: please add fillter here `examples | filter: search.value`

src/app/app.component.scss
```diff
@@ -37,4 +37,20 @@ h1.app-name {
 }
 .active {
     background-color: #efefef;
-}
\ No newline at end of file
+}
+.search {
+    margin: 15px;
+    width: 80%;
```
klumba12: make width 100% and adjust maring if needed

src/app/app.component.scss
```diff
@@ -37,4 +37,20 @@ h1.app-name {
...
+    display: block;
+     input{
+        height: 25px;
```
klumba12: why we need this?

src/app/app.component.scss
```diff
@@ -37,4 +37,20 @@ h1.app-name {
...
+    }
+}
+.mat-sidenav {
+   width: 260px;
```
klumba12: why?

PR #439, opened 2019-08-29
Feature/filter row

klumba12:


src/lib/assets/cell.scss
```diff
@@ -91,9 +91,9 @@ tbody {
             max-width: 80px;
         }
         &.q-grid-date {
-            width: 120px;
-            min-width: 110px;
-            max-width: 110px;
+            width: 140px;
```
etokareva: looks like value is the same. scss variable?

PR #484, opened 2020-07-20
9.1.6

Dalantren:


projects/qgrid-core/navigation/navigation.let.js
```diff
@@ -117,24 +120,35 @@ export class NavigationLet {
...
+							if (!td) {
+								td = table.body.cell(firstRowIndex, columnIndex).model();
+							}
+
+							if (!td) {
+								td = table.body.cell(rowIndex, firstColumnIndex).model();
+							}
+
+							if (!td) {
+								td = table.body.cell(firstRowIndex, firstColumnIndex).model();
```
etokareva, line 141: why did we need to repeat the same if condition for three times?

PR #498, opened 2020-11-17
feat(util/kit): handle cases when x or y is null

ddubrava:


projects/qgrid-core/utility/kit.js
```diff
@@ -48,15 +48,15 @@ const isEmail = value => {
 };
 
 function compare(x, y) {
-	if (x === y) {
-		return 0;
+	if (y === null || x > y) {
```
klumba12, line 51: I'm not sure about undefined also here, could you please google the default sort comparison function that is used in chrome, and copy it to here?

ddubrava: SortCompareDefault

- [2015](https://github.com/v8/v8/blob/950d2051a5ff065a5bc1d31f0e5d1bba850d0b3c/src/array.js#L898-L907)
- [Latest](https://github.com/v8/v8/blob/master/third_party/v8/builtins/array-sort.tq#L338-L364)

ddubrava: The implementation of $toString(x); is here [runtime.js](https://github.com/v8/v8/blob/950d2051a5ff065a5bc1d31f0e5d1bba850d0b3c/src/runtime.js#L803-L811). It's the default behaviour of `String()`;

But the problem is that converting null -> 'null' and undefined -> 'undefined' causes wrong sorting. So we should implement custom compare function which handles such values. I've created some examples - [sorting null and undefined](https://codepen.io/ddubrava/pen/ExyJybg?editors=0002)

klumba12 (approved):

PR #505, opened 2020-11-26
bugfix/unit_tests

ddubrava:
Not fixed:
- [group.let.spec.js](https://github.com/qgrid/ng2/blob/master/projects/qgrid-core/group/group.let.spec.js)
- [selection.service.spec.js](https://github.com/qgrid/ng2/blob/master/projects/qgrid-core/selection/selection.service.spec.js)


projects/qgrid-core/services/convert.spec.js
```diff
@@ -97,12 +97,13 @@ describe('Convert', () => {
 
 	describe('parseDate', () => {
```
klumba12, line 98: please rename parseDateTime and add testst for just parseDate

klumba12 (approved):

PR #511, opened 2020-12-10
WIP: added default label function to date and datetime fields

Dalantren:


projects/qgrid-core/column-type/date.column.js
```diff
@@ -21,6 +24,14 @@ export class DateColumnModel extends DataColumnModel {
 		super('date');
 
 		this.format = 'MM/dd/yyyy';
+
+		this.label = function (row) {
+			const value = getValue(row, this);
+			const parse = parseFactory('date');
+			const date = parse(value);
+			const isValidDate = getType(date) === 'datetime' && !isNaN(date);
```
klumba12, line 32: why we need this?

projects/qgrid-core/column-type/date.column.js
```diff
@@ -21,6 +24,14 @@ export class DateColumnModel extends DataColumnModel {
 		super('date');
 
 		this.format = 'MM/dd/yyyy';
+
+		this.label = function (row) {
+			const value = getValue(row, this);
+			const parse = parseFactory('date');
```
klumba12, line 30: please, move it from this function, it's not required to be inside

projects/qgrid-core/column-type/date.column.js
Dalantren: sometimes date parser could return an Invalid Date object (for not valid strings etc). In that case app crashes in date pipe

projects/qgrid-core/column-type/date.column.js
Dalantren: updated

projects/qgrid-core/column-type/date.column.js
klumba12: please use 	qGridDate pipe instead of angular date pipe, there you don't need to check on isValidDate

klumba12 (approved):

PR #518, opened 2021-02-08
Feature/dalantren/example pane with selection

Dalantren:


src/examples/pane-with-selection-mode/example-pane-with-selection-mode.component.html
```diff
@@ -0,0 +1,19 @@
...
+	<q-grid-pane>
+		<ng-template for="right" let-$pane let-$item="value">
+			<div style="width: 300px; display: flex;">
+				<h2>Discovered By: {{ $item?.discoveredBy }}</h2>
+				<button mat-icon-button (click)="$pane.close('right')">
+					<mat-icon>close</mat-icon>
+				</button>
+			</div>
+		</ng-template>
+	</q-grid-pane>
```
klumba12, line 18: add new line after

src/examples/pane-with-selection-mode/example-pane-with-selection-mode.component.ts
```diff
@@ -0,0 +1,43 @@
...
+	changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class ExamplePaneWithSelectionComponent implements AfterViewInit {
+	@ViewChild(GridComponent) grid: GridComponent;
+	@ViewChild(PaneComponent) pane: PaneComponent;
+
+	static tags = EXAMPLE_TAGS;
+	title = EXAMPLE_TAGS[1];
+
+	rows: Observable<Atom[]>;
```
klumba12, line 25: please rename it to rows$

src/examples/pane-with-selection-mode/example-pane-with-selection-mode.component.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	constructor(dataService: DataService) {
+		this.rows = dataService.getAtoms();
+	}
+
+	ngAfterViewInit(): void {
+		this.grid.model.mouseChanged.on(({ state }) => {
```
klumba12, line 32: can you make like, 

```typescript
const { model } = this.grid;
````

File: src/examples/pane-with-selection-mode/example-pane-with-selection-mode.component.ts [hunk 4]
```diff
@@ -0,0 +1,43 @@
...
+			const { code, status, target } = state;
+			if (code === 'left' && status === 'down') {
+				if (target && target.column.type !== 'select') {
+					this.pane.close('right');
+					this.pane.open('right');
+					Object.assign(this.pane.context, { value: target.row });
```
klumba12, line 38: can we make a better way to do it?

projects/ngx-qgrid-plugins/src/lib/pane/pane.component.ts
```diff
@@ -2,6 +2,8 @@ import { Component, OnInit, ChangeDetectorRef, Input, ChangeDetectionStrategy }
 import { GridPlugin, GridEventArg } from '@qgrid/ngx';
 import { TemplateHostService, GridError } from '@qgrid/ngx';
 
+export type PaneSide = 'left' | 'right';
```
klumba12, line 5: it won't work for the build as it's not exported from public_api

projects/ngx-qgrid-plugins/src/lib/pane/pane.component.ts
```diff
@@ -35,37 +39,43 @@ export class PaneComponent implements OnInit {
...
-						this.close('right');
-						this.open('right');
+						this.updateAndOpen('right')
 					}
 				});
 		}
 	}
 
-	open(side: 'right') {
+	open(side: PaneSide, value: any = null) {
```
klumba12, line 48: make just `value?: any` without null

projects/ngx-qgrid-plugins/src/lib/pane/pane.component.ts
```diff
@@ -35,37 +39,43 @@ export class PaneComponent implements OnInit {
...
 		const { table, model } = this.plugin;
 
-		let value = null;
 		const scope = this.parse();
-		if (scope) {
+		if (scope && !value) {
```
klumba12, line 52: should check `isUndefined(value)` because if the value is `0` or `false` that won't work

projects/ngx-qgrid-plugins/src/lib/pane/pane.component.ts
```diff
@@ -35,37 +39,43 @@ export class PaneComponent implements OnInit {
...
 	}
 
-	close(side: 'right') {
+	close(side: PaneSide) {
 		const { table } = this.plugin;
 
 		table.view.removeLayer(`pane-${side}`);
 
-		this.context = null;
+		this.context[side] = null;
```
klumba12, line 68: make a default for the side

projects/ngx-qgrid-plugins/src/lib/pane/pane.component.html
```diff
@@ -2,7 +2,14 @@
     <ng-template for="pane-right">
         <div class="q-grid-pane q-grid-pane-right">
             <ng-container key="plugin-pane-right.tpl.html"
-                          [context]="context">
+                          [context]="context?.right">
```
klumba12, line 5: this is a bad practice to use `?` if it's possible to have a default  - do it

projects/ngx-qgrid-plugins/src/lib/pane/pane.component.ts
```diff
@@ -35,37 +39,43 @@ export class PaneComponent implements OnInit {
...
 
+		this.invalidate();
+	}
+
+	updateAndOpen(side: PaneSide, value: any = null) {
```
klumba12, line 73: why we need that method? as I understood just call `open` should be enough in the example code?

Dalantren: nope. It will not re-render existing pane.
You had the same code previously in ngOnInit() on trigger
image

projects/ngx-qgrid-plugins/src/lib/pane/pane.component.ts
Dalantren: done

src/examples/pane-with-selection-mode/example-pane-with-selection-mode.component.html
Dalantren: done

src/examples/pane-with-selection-mode/example-pane-with-selection-mode.component.html
```diff
@@ -0,0 +1,21 @@
+<q-grid [caption]="title"
+		selectionUnit="row"
+		selectionMode="multiple"
+		selectionArea="custom"
+		[rows]="rows$ | async">
```
klumba12, line 5: still not correct format?

klumba12 (approved):

PR #526, opened 2021-03-10
row min height when resizing

regeldso:
added property for setting up row minimum height

added condition for row min height validation

where to check: examples -> size-row-basic -> change row size below minimal value

### DIFF
--- a/projects/qgrid-core/layout/layout.let.js
+++ b/projects/qgrid-core/layout/layout.let.js
@@ -3,7 +3,6 @@ import * as css from '../services/css';
 import * as columnService from '../column/column.service';
 import { Fastdom } from '../services/fastdom';
 import { selectColumn } from '../navigation/navigation.state.selector';
-import { RowState } from '../row/row.state';
 
 export class LayoutLet {
 	constructor(plugin) {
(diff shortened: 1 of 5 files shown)


projects/qgrid-core/layout/layout.let.js
```diff
@@ -166,6 +167,10 @@ export class LayoutLet {
 		const form = layout().rows;
 		const style = form.get(row);
 		if (style) {
+			const rowState = model.row();
+			if (rowState.minHeight > style.height){
+				return;
+			}
 			context.class(`resized-${style.height}px`, { height: style.height + 'px' });
```
klumba12, line 174: add a new line after row #174

projects/qgrid-core/layout/layout.let.js
```diff
@@ -3,6 +3,7 @@ import * as css from '../services/css';
 import * as columnService from '../column/column.service';
 import { Fastdom } from '../services/fastdom';
 import { selectColumn } from '../navigation/navigation.state.selector';
+import { RowState } from '../row/row.state';
```
klumba12, line 6: remove this import

projects/qgrid-core/layout/layout.let.js
```diff
@@ -166,6 +167,10 @@ export class LayoutLet {
 		const form = layout().rows;
 		const style = form.get(row);
 		if (style) {
+			const rowState = model.row();
```
klumba12, line 170: consider to use deconstruction
```const { minHeight } = model.row();```

klumba12 (approved):

PR #579, opened 2022-03-15
add class for q-grid element if layer exists

gwhrus:
If q-grid has q-grid-layer, add class 'q-grid-${name}' to q-grid element while layer is present


packages/qgrid-ngx/src/lib/layer/layer.service.ts
```diff
@@ -25,6 +25,13 @@ export class LayerService {
 			const { nativeElement } = container.element;
 			nativeElement.parentElement.classList.add(`q-grid-layer-${name}`);
 
+		for (let el = nativeElement; !!el; el = el.parentElement) {
```
klumba12, line 28: consider to put the logic of finding q-grid element under the sep. function like getHost(): HTMLElement {}

gwhrus: fixed

klumba12 (approved):

PR #580, opened 2022-03-17
row number column fixed

gwhrus:
row number column: incorrect numbers shown on pages starting from 2nd,
fixed template and pagination getters

### DIFF
--- a/packages/qgrid-core/src/pagination/pagination.let.js
+++ b/packages/qgrid-core/src/pagination/pagination.let.js
@@ -2,6 +2,8 @@ export class PaginationLet {
 	constructor(plugin) {
 		const { model, observe } = plugin;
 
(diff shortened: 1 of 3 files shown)


packages/qgrid-core/src/pagination/pagination.let.js
```diff
@@ -2,6 +2,8 @@ export class PaginationLet {
 	constructor(plugin) {
 		const { model, observe } = plugin;
 
+		Object.defineProperty(this, 'current', { get: () => plugin.model.pagination().current });
```
klumba12, line 5: Please just do like this.plugin = plugin; consider not to use Object.defineProperty, we are on es6 :)

klumba12 (approved):

PR #582, opened 2022-03-17
Mandatory column chooser

FinalButterfly:
- Column chooser now present by default in actions array
- Actions will be sorted by priority field in their Command object

### DIFF
--- a/packages/qgrid-core/src/command/command.d.ts
+++ b/packages/qgrid-core/src/command/command.d.ts
@@ -1,5 +1,5 @@
-import { CommandContext } from './command.context';
 import { SubjectLike } from '../rx/rx';
+import { CommandContext } from './command.context';
 
 /**
  * Use this class to implement command pattern in the qgrid. The most of interactions in the q-grid are utilized by this pattern.
@@ -28,7 +28,7 @@ import { SubjectLike } from '../rx/rx';
  * * [Command Pattern Wiki](https://en.wikipedia.org/wiki/Command_pattern)
  */
 export declare class Command<T = any> {
-    constructor(context?: CommandContext<T>);
+    constructor(context?: CommandContext<T>, priority?: number);
 
     /**
      * Indicates if a command can be invoked. Use one argument to support typescript generic typification.
--- a/packages/qgrid-core/src/command/command.d.ts
+++ b/packages/qgrid-core/src/command/command.d.ts
@@ -28,7 +28,7 @@ import { CommandContext } from './command.context';
  * * [Command Pattern Wiki](https://en.wikipedia.org/wiki/Command_pattern)
  */
 export declare class Command<T = any> {
-    constructor(context?: CommandContext<T>, priority?: number);
+    constructor(context?: CommandContext<T>);
 
     /**
      * Indicates if a command can be invoked. Use one argument to support typescript generic typification.
--- a/packages/qgrid-core/src/command/command.js
+++ b/packages/qgrid-core/src/command/command.js
@@ -1,14 +1,14 @@
-import { yes } from '../utility/kit';
 import { SubjectLike } from '../rx/rx';
+import { yes } from '../utility/kit';
 
 export class Command {
-	constructor(context = {}) {
+	constructor(context = {}, priority = 1) {
 		this.execute = yes;
 		this.canExecute = yes;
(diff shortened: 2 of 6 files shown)


packages/qgrid-core/src/command/command.d.ts
```diff
@@ -28,7 +28,7 @@ import { SubjectLike } from '../rx/rx';
  * * [Command Pattern Wiki](https://en.wikipedia.org/wiki/Command_pattern)
  */
 export declare class Command<T = any> {
-    constructor(context?: CommandContext<T>);
+    constructor(context?: CommandContext<T>, priority?: number);
```
klumba12, line 31: why priority is not a part of CommandContext?

packages/qgrid-ngx-plugins/src/lib/action/action.component.ts
```diff
@@ -43,10 +43,14 @@ export class ActionComponent implements OnInit {
 				.action()
 				.items
 				.concat([action])
+				.sort((a1: Action, a2: Action) => {
```
klumba12, line 46: typically in sort functiona arguments called like `a` and `b`

packages/qgrid-core/src/command/command.js
```diff
@@ -1,14 +1,14 @@
-import { yes } from '../utility/kit';
 import { SubjectLike } from '../rx/rx';
+import { yes } from '../utility/kit';
 
 export class Command {
-	constructor(context = {}) {
+	constructor(context = {}, priority = 1) {
```
klumba12, line 5: let's keep priority -1 as a default (it will mean no priority), put prioirty under context please

packages/qgrid-ngx-plugins/src/lib/action-bar/action-bar.component.ts
```diff
@@ -25,6 +25,24 @@ export class ActionBarComponent implements OnInit {
 
 		observe(model.actionChanged)
 			.subscribe(e => {
+				const initialItems = e.state.items;
+				const sortedItems = initialItems.sort((a: Action, b: Action) => {
+					return a.command.priority - b.command.priority;
+				});
+
+				const sorted = sortedItems.every((item: Action, index: number) => {
+					return item.title = initialItems[index].title;
```
klumba12, line 34: = or === ?

packages/qgrid-ngx-plugins/src/lib/action/action.component.ts
```diff
@@ -43,10 +43,14 @@ export class ActionComponent implements OnInit {
 				.action()
 				.items
 				.concat([action])
+				.sort((a: Action, b: Action) => {
```
klumba12, line 46: I don't think that we need that code

packages/qgrid-ngx-plugins/src/lib/action-bar/action-bar.component.ts
klumba12: maybe a good idea to check if array is sorted first, and if not sort it out?

https://stackoverflow.com/questions/19458278/check-if-an-array-is-sorted-return-true-or-false

packages/qgrid-ngx-plugins/src/lib/action-bar/action-bar.component.ts
```diff
@@ -36,4 +50,14 @@ export class ActionBarComponent implements OnInit {
 		const { model } = this.plugin;
 		return model.action().items;
 	}
+
+	isSorted(array: Action[]): boolean {
```
klumba12, line 54: make it private, consider to rename array -> actions

packages/qgrid-ngx-plugins/src/lib/action-bar/action-bar.component.ts
```diff
@@ -36,4 +50,14 @@ export class ActionBarComponent implements OnInit {
...
+		for (let i = 0; i < array.length - 1; i++) {
+			if (array[i].command.priority > array[i + 1].command.priority) {
```
klumba12, line 56: will be a little bit more readable if do:
const action = actions[i];
const nextAction = actions[i + 1];

klumba12 (approved):

PR #587, opened 2022-03-25
Enabled row expansion on double click

FinalButterfly:


packages/qgrid-ngx/src/lib/row/row.component.ts
```diff
@@ -90,6 +90,31 @@ export class RowComponent implements OnChanges, OnInit {
...
 			}
 		}
+
+
+		let firstClickTimestamp = null;
+		let targetTd = null;
+
+		observe(model.mouseChanged)
+			.subscribe(e => {
+				const { code, status, target, timestamp } = e.state;
```
klumba12, line 100: why not to use e.changes?

```typescript
                  const {timestamp} = e.changes;
                  const isDoubleclick = (timestamp.newValue – timestamp.oldValue) < 300; // < 300ms
```

klumba12 (approved):

PR #595, opened 2022-04-18
Added basic eslint config files to extend

Dalantren:


.eslintrc.json
```diff
@@ -0,0 +1,81 @@
...
+      "rules": {
+        "arrow-body-style": "error",
+        "brace-style": [
+          "error",
+          "1tbs",
+          { "allowSingleLine": true }
+        ],
+        "constructor-super": "error",
+        "curly": "error",
+        "dot-notation": "off",
```
FinalButterfly, line 22: We should enable this.

`field.property` is preferable to `field["property"]`.

FinalButterfly (approved):

.eslintrc.json
klumba12: not always

klumba12 (approved):

